PR #30, opened 2020-06-01
Add ability to check update and delete policies outside of mutation

### LINKED ISSUE #21: Add ability to run update/delete privacy policy outside of mutator
This is useful for situations where putting a sequence of mutations in a transaction isn't possible, and thus a reasonable effort must be made ahead of time of the mutations to guess as to whether they would fail.

For example:
```typescript
runInTransaction(() => {
  const someThirdPartyServiceID = await StripedZebra.makePaymentAsync(info);
  await PurchaseEntity.creator(viewerContext).setField('zebra_id', someThirdPartyServiceID).createAsync(); // this could fail due to privacy policy
});
```

In this case, there's no way to undo the third party API call when the entity creation privacy policy fails.

Something like the following seems to work. Would be useful to clean up and formalize into a sensible API.

```typescript
static async canViewerUpdate<
    TMFields,
    TMID,
    TMViewerContext extends ViewerContext,
    TMEntity extends Entity<TMFields, TMID, TMViewerContext>,
    TMPrivacyPolicy extends EntityPrivacyPolicy<TMFields, TMID, TMViewerContext, TMEntity>
  >(
    this: IEntityClass<TMFields, TMID, TMViewerContext, TMEntity, TMPrivacyPolicy>,
    existingEntity: TMEntity,
    queryContext: EntityQueryContext = existingEntity
      .getViewerContext()
      .getViewerScopedEntityCompanionForClass(this)
      .getQueryContextProvider()
      .getRegularEntityQueryContext()
  ): Promise<boolean> {
    const privacyPolicy = new (this.getCompanionDefinition().privacyPolicyClass)();
    const evaluationResult = await asyncResult(
      privacyPolicy.authorizeUpdateAsync(
        existingEntity.getViewerContext(),
        queryContext,
        existingEntity
      )
    );
    return evaluationResult.ok;
  }
```

wschurman:
# Why

Closes #21. 

# How

Add a set of new static methods that we've found would be useful in expo server.

# Test Plan

Run new tests.

### DIFF
--- a/packages/entity/src/Entity.ts
+++ b/packages/entity/src/Entity.ts
@@ -1,4 +1,4 @@
-import { Result } from '@expo/results';
+import { Result, asyncResult } from '@expo/results';
 
 import { EntityCompanionDefinition } from './EntityCompanionProvider';
 import { CreateMutator, UpdateMutator } from './EntityMutator';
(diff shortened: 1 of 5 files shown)


packages/entity/src/Entity.ts
```diff
@@ -140,6 +140,83 @@ export default abstract class Entity<
...
+  /**
+   * Check whether an entity loaded by a viewer can be deleted by that same viewer.
+   *
+   * @remarks
+   * See remarks for canViewerUpdate.
+   *
+   * @param existingEntity - entity loaded by viewer
+   * @param queryContext - query context in which to perform the check
+   */
+  static async canViewerDelete<
```
ide, line 195: -Async

wschurman: Ah nice catch!

PR #32, opened 2020-06-02
Fix async method naming in Entity

wschurman:
# Why

https://github.com/expo/entity/pull/30#pullrequestreview-423003224

# Test Plan

`yarn tsc`


wschurman:
https://github.com/expo/entity/issues/33

PR #46, opened 2020-06-17
feat!: add database adapter provider for better injection

wschurman:
# Why

- This brings database adapter inject-ability up to par with cache adapters and query contexts.
- This will allow unit tests to more appropriately test logic. In the particular case that sparked this PR, this will allow unit tests to test two entities reading and writing to the same table which is a use case we've stumbled upon at expo.

# How

1. Change injection pattern to match that of cache adapters.
1. Update tests to support the injection of raw data rather than object data.
1. Add database adapter providers for postgres.

# Test Plan

`yarn test`
`yarn integration`


packages/entity/src/IEntityDatabaseAdapterProvider.ts
```diff
@@ -0,0 +1,15 @@
+import EntityConfiguration from './EntityConfiguration';
+import EntityDatabaseAdapter from './EntityDatabaseAdapter';
+
+/**
+ * A database adapter provider vends database adapters for a particular database adapter type.
+ * Allows for passing global configuration to databse adapters, making testing easier.
```
jkhales, line 6: Ah I see.

PR #47, opened 2020-06-18
feat!: add support for entities sharing the same table

wschurman:
# Why & How

The current implementation of entities had a few caching problems when trying to do this. Using the following as an example:
- Entity class EntityA: `id` is cached, table is `blah`, type is A, otherField string
- Entity class EntityB: `id` is cached, table is `blah`, type is B
- Entity instance 1: `id = 1`, `type = A`
- Entity instance 2: `id = 2`, `type = B`

1. "Invalid loads" would be cached if loaded by a field that is cached. Let's say I did a `EntityA.loader(vc).loadByID(2)`. This would load successfully from the DB, and be cached. Then, when returned to the user the business logic would see that the type is incorrect and would throw, but at this point it is too late as the entity is already in the dataloader and cache and only contains fields from A. To solve this, the validation logic is moved to happen at the database adapter level so that if an "invalid" load is encountered it acts as if no data was returned for that query.
1. "Invalid loads" would negatively cache invalid rows for the cache key, which could be the same for both entities. Let's say I did a `EntityA.loader(vc).loadByID(2)`. The new validator from (1) would nullify the load, and the cache adapter would treat this as a MISS, thus negatively caching the key value. To solve this, create a mechanism for having separate caches even with the same backing table.

# Test Plan

Run tests (old and new).

### DIFF
--- a/packages/entity-cache-adapter-redis/src/RedisCacheAdapter.ts
+++ b/packages/entity-cache-adapter-redis/src/RedisCacheAdapter.ts
@@ -156,10 +156,7 @@ export default class RedisCacheAdapter<TFields> extends EntityCacheAdapter<TFiel
   ): string {
     return this.context.makeKeyFn(
       this.context.cacheKeyPrefix,
-      this.entityConfiguration.tableName,
-      `v${this.entityConfiguration.cacheKeyVersion}`,
-      fieldName as string,
-      String(fieldValue)
+      ...this.getCacheKeyParts(fieldName, fieldValue)
     );
   }
 }
--- a/packages/entity/src/EntityCacheAdapter.ts
+++ b/packages/entity/src/EntityCacheAdapter.ts
@@ -7,7 +7,7 @@ import { CacheLoadResult } from './internal/ReadThroughEntityCache';
  * cached, fetched from cache, and removed from cache (invalidated).
  */
 export default abstract class EntityCacheAdapter<TFields> {
-  constructor(protected readonly entityConfiguration: EntityConfiguration<TFields>) {}
+  constructor(private readonly entityConfiguration: EntityConfiguration<TFields>) {}
 
   /**
    * Transformer definitions for field types. Used to modify values as they are read from or written to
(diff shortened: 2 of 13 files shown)


ide:
The main question I think this diff is partially answering is whether we want the Redis cache to be per table (a faster way to access the database) or per entity (a faster way to load an entity), and this diff chooses the latter.

If we wanted to implement the former, we wouldn't cache "invalid" entries at all. We'd always load the data from Redis (or Postgres if Redis is empty) and filter out invalid entries at that level.

The pros of a database cache are that we have one cache per table, so if two entities wanted to read overlapping rows (if users and robots were not mutually exclusive, for example), we'd cache those overlapping rows only once instead of once per entity type. The database cache also make cache invalidation for a table easier - there's just one cache to consider.

The pros of an entity cache are that we can flush the cache for just one entity without disrupting the others. We also can cache entity-specific transformations of the DB rows instead of the raw DB data -- "invalid" rows are "transformed" to MISSes in this PR.

My sense is that we should think a few steps ahead before making a decision. How do these caching strategies affect dependent entities (I think it doesn't matter, but something to consider). How do they affect transactions if we were to build transaction-aware caching?

wschurman:
> The pros of a database cache are that we have one cache per table, so if two entities wanted to read overlapping rows (if users and robots were not mutually exclusive, for example), we'd cache those overlapping rows only once instead of once per entity type. The database cache also make cache invalidation for a table easier - there's just one cache to consider.

The issue here is that different entities can currently define different sets of fields on the same table, and can define different cache key versions as well, meaning the framework is already a bit more skewed to the "per entity" route you mention. 

In the current state of the world, when an entity is loaded from the table, unspecified fields are ignored. If we were to decide to go down this route, we'd instead want to somehow make it possible to create a "derived entity", where we'd keep the current setup and actually enforce that there's only one entity per table. Then, we could have "derived entities" that somehow are a view on the table entity, only exposing a subset of the table entity's fields. Or, we could create a new concept that enforces a single `EntityDataManager` per table and the fields would be defined alongside that, but then multiple entities share the `EntityDataManager` and an entity definition is then a view on the fields (I like this route actually, I'll create a new PR exploring this route).

> My sense is that we should think a few steps ahead before making a decision. How do these caching strategies affect dependent entities (I think it doesn't matter, but something to consider). How do they affect transactions if we were to build transaction-aware caching?

The gut sense I get is that we need to decide whether we want this feature (multiple entities backed by a single table) to be:
- a first-class feature that drives the architecture of the framework as a whole (how entities, fields, tables, etc are defined).
- a narrow special-case that doesn't change the current setup too much, but is also fairly non-obvious how to accomplish it.

wschurman:
Put up an alternative PR to this that explores the concept keeping entity data operation on a table level (as it currently is in prod), and just modifying the behavior of entity to be a "view" into that data. https://github.com/expo/entity/pull/49

I actually am leaning towards that one for one reason: it's less moving parts and simpler conceptually. I detailed the pros and cons on the PR though.

wschurman:
Abandoning this in favor of #49 which satisfies two known use cases better:
- Overlapping rows, where two different entities can be loaded from the same ID.
- Disjoint rows, where two different entities are backed by the same table but have mutually exclusive rows.

PR #51, opened 2020-06-29
Change key specification to extends

wschurman:
# Why

An experiment to see if we can get this to typecheck


wschurman:
Closing this as it still doesn't feel quite as clean as just specifying the set of fields. Definitely may be a future area of development though.

PR #65, opened 2020-07-22
feat: mutation triggers

wschurman:
# Why

This PR adds mutation triggers, which are similar to [active record callbacks](https://guides.rubyonrails.org/active_record_callbacks.html). They allow entity authors to specify code that runs before and after mutations, and a special callback that occurs after commit of the transaction.

Use cases:
- Audit logging (similar to https://github.com/paper-trail-gem/paper_trail)
- Third-party service calls to keep data in sync
- Email sending
- etc...

Note that for now, this is separate from full entity validation (not yet implemented), but that will be done in a similar way. Depending on the transaction semantics we want for validation it may be exactly the same as `beforeAll` but it's unclear.

# How

Add ability to specify features, add tests.

# Test Plan

Run tests.

### DIFF
--- a/packages/entity/src/EntityMutationTrigger.ts
+++ b/packages/entity/src/EntityMutationTrigger.ts
@@ -1,0 +1,77 @@
+import { EntityQueryContext } from './EntityQueryContext';
+import ReadonlyEntity from './ReadonlyEntity';
+import ViewerContext from './ViewerContext';
+
+/**
+ * Interface to define trigger behavior for entities.
+ */
+export interface EntityMutationTriggerConfiguration<
+  TFields,
+  TID,
+  TViewerContext extends ViewerContext,
+  TEntity extends ReadonlyEntity<TFields, TID, TViewerContext, TSelectedFields>,
+  TSelectedFields extends keyof TFields = keyof TFields
+> {
+  /**
+   * Trigger set that runs within the transaction but before the entity is created in the database.
+   */
+  beforeCreate?: EntityMutationTrigger<TFields, TID, TViewerContext, TEntity, TSelectedFields>[];
+  /**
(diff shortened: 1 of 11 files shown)


packages/entity/src/EntityMutationTrigger.ts
```diff
@@ -0,0 +1,77 @@
...
+  TViewerContext extends ViewerContext,
+  TEntity extends ReadonlyEntity<TFields, TID, TViewerContext, TSelectedFields>,
+  TSelectedFields extends keyof TFields = keyof TFields
+> {
+  abstract async executeAsync(
+    viewerContext: TViewerContext,
+    queryContext: EntityQueryContext,
+    entity: TEntity
+  ): Promise<void>;
+}
```
ide, line 77: Unless there's a meaningful reason to make this a class, plain async functions would be more lightweight:
```ts
type EntityMutationTrigger<
    TFields,
    TID,
    TEntity extends ReadonlyEntity<TFields, TID, TViewerContext, TSelectedFields>,
    TSelectedFields extends keyof TFields = keyof TFields
> = (
      viewerContext: TViewerContext,
      queryContext: EntityQueryContext,
      entity: TEntity
    ) => Promise<void>;
```

and the replacement for the constructor looks like:

```ts
function createThrowConditionallyTrigger<...>(parameters): EntityMutationTrigger<...> {
  return async function throwConditionallyAsync() {
    ... close over parameters if needed
  }
}
```

wschurman: So far the class approach has worked out fairly cleanly for `EntityPrivacyPolicyRule`, especially since feeding params into the constructor is common as is having overloaded constructors. I can imagine that these may turn out to be reasonably similar, but I'll definitely keep it in mind as we start writing triggers and seeing usage patterns.

Issue #25, opened 2020-05-30
Field Validation


wschurman:
Entity fields should do basic validation.
- UUIDField should validate that it's a UUID
- Primitive fields should validate their types
- EnumField should validate value being a member of a specified enum
- Custom validation functions should be possible as well.

wschurman:
These could do validation on `setField` (or `createAsync` if we want to do validation of all the set fields at once), `loadByField`, and we'll also want to decide whether this should just be for those APIs or also at load time (coming from the DB).

wschurman:
This was done a while ago.

PR #67, opened 2020-07-23
feat: entity mutation validators

quinlanj:
# Why
Implements validation for entity mutations. Validators are specified by the user, and run before every create and update mutation. Validators are run before any triggers.

https://github.com/expo/entity/issues/25

# Tests

- [x] new unit tests pass (verifying validator call counts on different mutation operations)
- [x] new integration tests pass (verifying validators work as expected inside a txn)

### DIFF
--- a/packages/entity/src/EntityMutationTrigger.ts
+++ b/packages/entity/src/EntityMutationTrigger.ts
@@ -2,19 +2,6 @@ import { EntityQueryContext } from './EntityQueryContext';
 import ReadonlyEntity from './ReadonlyEntity';
 import ViewerContext from './ViewerContext';
 
-export interface EntityMutationValidatorConfiguration<
-  TFields,
-  TID,
-  TViewerContext extends ViewerContext,
-  TEntity extends ReadonlyEntity<TFields, TID, TViewerContext, TSelectedFields>,
-  TSelectedFields extends keyof TFields = keyof TFields
-> {
-  beforeCreate?: EntityMutationTrigger<TFields, TID, TViewerContext, TEntity, TSelectedFields>[];
-  beforeUpdate?: EntityMutationTrigger<TFields, TID, TViewerContext, TEntity, TSelectedFields>[];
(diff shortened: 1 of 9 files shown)


packages/entity/src/EntityMutationTrigger.ts
```diff
@@ -2,6 +2,19 @@ import { EntityQueryContext } from './EntityQueryContext';
...
+  TFields,
+  TID,
+  TViewerContext extends ViewerContext,
+  TEntity extends ReadonlyEntity<TFields, TID, TViewerContext, TSelectedFields>,
+  TSelectedFields extends keyof TFields = keyof TFields
+> {
+  beforeCreate?: EntityMutationTrigger<TFields, TID, TViewerContext, TEntity, TSelectedFields>[];
+  beforeUpdate?: EntityMutationTrigger<TFields, TID, TViewerContext, TEntity, TSelectedFields>[];
+  beforeDelete?: EntityMutationTrigger<TFields, TID, TViewerContext, TEntity, TSelectedFields>[];
+  beforeAll?: EntityMutationTrigger<TFields, TID, TViewerContext, TEntity, TSelectedFields>[];
```
wschurman, line 15: Thinking about this more, there should only be one set of validators and it should run on create and update. This matches active record https://guides.rubyonrails.org/active_record_validations.html

Because of this, I'd say we can just get rid of `EntityMutationValidatorConfiguration` and just have the mutation validators field be an array: `mutationValidators: EntityMutationTrigger<TFields, TID, TViewerContext, TEntity, TSelectedFields>[];`

packages/entity/src/__tests__/EntityMutator-test.ts
```diff
@@ -185,6 +185,18 @@ const createEntityMutatorFactory = (
     keyof TestFields
   >;
 } => {
+  const mutatationValidators: EntityMutationTriggerConfiguration<
```
wschurman, line 188: `EntityMutationValidatorConfiguration`

PR #68, opened 2020-07-23
fix: separate out EntityMutationValidator

wschurman:
# Why

Now that these are used for both triggers and validators, they should have a distinct name. I considered the following, but am open to others suggestions as well:
- `EntityMutationCallback` - callbacks usually don't prevent something from running, so it didn't make sense for validation
- `EntityMutationHandler` - handlers generally refer to overriding behavior of the framework rather than hooking into it.
- `EntityMutationExecutable`- the method inside it is called `executeAsync` so this was the natural name
- `EntityMutationLambda` - not really a JS term
- `EntityMutationHook` - The term `hook` refers more to the when it is run rather than the what is run IMO.

# How

TS renames.

# Test Plan

`yarn tsc`


packages/entity/src/EntityMutationExecutable.ts
```diff
@@ -0,0 +1,25 @@
+import { EntityQueryContext } from './EntityQueryContext';
+import ReadonlyEntity from './ReadonlyEntity';
+import ViewerContext from './ViewerContext';
+
+/**
+ * A trigger is a way to specify entity mutation operation side-effects that run within the
+ * same transaction as the mutation itself. The one exception is afterCommit, which will run within
+ * the transaction if a transaction is supplied.
+ *
+ * A validator is a way to specify entity mutation validation that run within the
```
ide, line 10: run*s*

wschurman:
That's good reasoning for separating out the types. I updated the PR to reflect that.

As for the class vs function, see this comment: https://github.com/expo/entity/pull/65#discussion_r459184745. I'll definitely consider making it a function in the future if usage seems to not make use of the benefits of a class, but having it as a class has turned out to be reasonably helpful for privacy policy rules.

Issue #76, opened 2020-08-18
Investigate distributed reader-writer locks for entities


wschurman:
To guarantee cache consistency (and dataloader consistency) across a set of machines or even across requests and keep the read-through caching strategy that we employ, we'll need to implement some sort of distributed reader-writer lock for entities by type and ID.

Another interesting way to mitigate this temporarily but not theoretically fix it is to defer all cache invalidation to the end of the write transaction so that it becomes less likely to have a read-through cache inconsistency (though not impossible still obviously).

wschurman:
This is [documented](https://github.com/expo/entity/blob/main/packages/entity/src/AuthorizationResultBasedEntityMutator.ts#L36) at the  and somewhat solved (as best it can be) without reader-writer locks.

PR #77, opened 2020-08-19
fix: move cache invalidation to after transaction commit

wschurman:
# Why

Implementation of second idea from #76.

# How

- Append bound async functions to array of fns that are executed after transaction has been committed. 
- Move general entity integration tests into separate package that integrate all cache and database adapters.
- Write a test that fails with the previous implementation and passes with the new implementation (`EntityCacheInconsistency-test`). This test runs an update in a long-running transaction and does a single targeted load between when the cache is invalidated (in the old implementation) and when the transaction completes. With the old implementation, the cache would have been cleared before the change was visible to other machines. With the new implementation, the cache is cleared at commit time (right after commit). Technically there still could be a race if the loads of a single entity are super frequent and an update occurs but that problem can't be solved without a distributed reader-writer lock which at this point is probably overkill.

# Test Plan

Run new test.


wschurman:
One thing to figure out is if we want these invalidations to occur right before or right after the transaction commit.

packages/entity/src/EntityQueryContext.ts
```diff
@@ -34,23 +36,25 @@ export class EntityNonTransactionalQueryContext extends EntityQueryContext {
...
-      return await transactionScope(this);
-    } else {
-      return await this.entityQueryContextProvider.runInTransactionAsync(transactionScope);
-    }
+    return await this.entityQueryContextProvider.runInTransactionAsync(transactionScope);
   }
 }
 
 export class EntityTransactionalQueryContext extends EntityQueryContext {
+  public readonly postCommitCallbacks: PostCommitCallback[] = [];
```
ide, line 46: API design discussion: once a post-commit callback is run, I don't think we ever want to run it again unless we guarantee the callbacks are idempotent. So I wonder if it makes sense to design this post-commit callback API in a way where running the callbacks clears them out so they are less likely to be run twice. Ex:

```ts
getPostCommitCallbacks(): PostCommitCallback[] {
  const callbacks = this.postCommitCallbacks;
  this.postCommitCallbacks.length = 0;
  return callbacks;
}
```
or
```ts
async runPostCommitCallbacksAsync(): Promise<void> {
  const callbacks = this.postCommitCallbacks;
  this.postCommitCallbacks.length = 0;
  await Promise.all(callbacks.map(callback => callback()));
}
``` 
depending on whether the caller of the query context should be responsible for running the callbacks and if the query context should just be responsible for holding onto some state.

packages/entity-full-integration-tests/src/__integration-tests__/EntityCacheInconsistency-test.ts
```diff
@@ -0,0 +1,190 @@
...
+  });
+
+  test('lots of updates in long-ish running transactions', async () => {
+    const viewerContext = new ViewerContext(
+      createFullIntegrationTestEntityCompanionProvider(knexInstance, redisCacheAdapterContext)
+    );
+
+    const entity1 = await TestEntity.creator(viewerContext)
+      .setField('other_string', 'hello')
+      .setField('third_string', 'wat')
```
ide, line 144: give this value something more descriptive like "initial value" or "precommit"

packages/entity-full-integration-tests/src/__integration-tests__/EntityCacheInconsistency-test.ts
```diff
@@ -0,0 +1,190 @@
...
+      (async () => {
+        await new Promise((r) => setTimeout(r, 100));
+        const viewerContextInternal = new ViewerContext(
+          createFullIntegrationTestEntityCompanionProvider(knexInstance, redisCacheAdapterContext)
+        );
+        await TestEntity.loader(viewerContextInternal).enforcing().loadByIDAsync(entity1.getID());
+      })(),
+      TestEntity.runInTransactionAsync(viewerContext, async (queryContext) => {
+        await TestEntity.updater(entity1, queryContext)
+          .setField('third_string', 'huh')
```
ide, line 167:
```suggestion
          .setField('third_string', 'postcommit')
```

packages/entity-full-integration-tests/src/__integration-tests__/EntityCacheInconsistency-test.ts
```diff
@@ -0,0 +1,190 @@
...
+      .enforcing()
+      .loadByFieldEqualingAsync('other_string', 'hello');
+
+    await Promise.all([
+      // do a load after the transaction below updates the entity but before transaction commits to ensure
+      // that the cache is cleared after the transaction commits rather than in the middle where the changes
+      // may not be visible by other requests (ViewerContexts) which would cache the incorrect
+      // value during the read-through cache.
+      (async () => {
+        await new Promise((r) => setTimeout(r, 100));
```
ide, line 159: For controlling the exact order of events I find explicit barriers to be useful. Something simple like:

```ts
let openBarrier1;
const barrier1 = new Promise((resolve) => {
  openBarrier1 = resolve;
});

...

// one task
await barrier1;

...

// the other task
openBarrier1();
```

This way the tests run as fast as possible and the desired order of events is explicitly expressed instead of implicitly with timeout values.

packages/entity/src/EntityQueryContextProvider.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  /**
+   * Get the query interface for constructing a query context.
+   */
+  protected abstract getQueryInterface(): any;
+
+  /**
+   * Vend a transaction runner for use in {@link runInTransactionAsync}.
+   */
+  protected abstract createTransactionRunner<T>(): (
```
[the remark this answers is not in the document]
wschurman: Not an async method (it vends an async method). Will keep as-is.

packages/entity/src/EntityQueryContext.ts
wschurman: Good suggestion! I'll modify it a bit to make all callback internals private to that class.

PR #83, opened 2020-10-08
feat: add entityClassName to metrics adapter events

wschurman:
# Why

This adds the ability for concrete implementations of `IEntityMetricsAdapter` to segregate metric records by entity type. While not required (implementors can do whatever they like with the events), it may be helpful to find things like entity hot paths, slow tables, places that could use caches, etc.

This is not a breaking change as the interface is an object so code that worked with the old interface should work with the new interface.

# How

Add parameter to definition and pipe through to log events.

# Test Plan

`yarn test` (updated tests test the new behavior in addition to the old behavior to ensure no regressions).

### DIFF
--- a/packages/entity/src/metrics/EntityMetricsUtils.ts
+++ b/packages/entity/src/metrics/EntityMetricsUtils.ts
@@ -6,7 +6,8 @@ import IEntityMetricsAdapter, {
 
 export const timeAndLogLoadEventAsync = (
   metricsAdapter: IEntityMetricsAdapter,
-  loadType: EntityMetricsLoadType
+  loadType: EntityMetricsLoadType,
+  entityClassName: string
 ) => async <TFields>(promise: Promise<readonly Readonly<TFields>[]>) => {
   const startTime = Date.now();
   const result = await promise;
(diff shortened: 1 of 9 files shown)


packages/entity/src/metrics/EntityMetricsUtils.ts
```diff
@@ -44,14 +48,16 @@ export const timeAndLogLoadMapEventAsync = (
 
 export const timeAndLogMutationEventAsync = (
   metricsAdapter: IEntityMetricsAdapter,
-  mutationType: EntityMetricsMutationType
+  mutationType: EntityMetricsMutationType,
+  entityClassName: string
```
ide, line 52: Just wanted to ask whether ordering this by `name, type` would make more sense than `type, name` depending on what you think is more important to put first

ide (approved):

packages/entity/src/metrics/EntityMetricsUtils.ts
wschurman: I thought briefly about this as I was adding it and chose `type, name` for the somewhat arbitrary reason that I liked the stricter type to precede the less strict type. That being said, I don't really have a preference unless we were to add another string argument, in which case I like to split them up with concrete types to make incorrect ordering less likely.

wschurman: Based on the fact that we added type before class, my gut says type is more important.

PR #89, opened 2020-11-04
fix: edge deletion behavior cyclic import structure

wschurman:
# Why

The way this was created led to import cycles in top level module exports, causing them to resolve as undefined. (as described in https://medium.com/visual-development/how-to-fix-nasty-circular-dependency-issues-once-and-for-all-in-javascript-typescript-a04c987cf0de)

# How

Make one end of the reference cycle a function to prevent necessary resolution at module import time.

# Test Plan

1. Move integration test entities to different files in the first commit. Run `yarn integration`, ensure they fail with the following:

```
TypeError: Cannot read property 'getCompanionDefinition' of undefined

      634 |       inboundEdges.map(async (entityClass) => {
      635 |         return await mapMapAsync(
    > 636 |           entityClass.getCompanionDefinition().entityConfiguration.schema,
          |                       ^
      637 |           async (fieldDefinition, fieldName) => {
      638 |             const association = fieldDefinition.association;
      639 |             if (!association) {

      at ../entity/src/EntityMutator.ts:636:23
          at Array.map (<anonymous>)
      at Function.processEntityDeletionForInboundEdgesAsync (../entity/src/EntityMutator.ts:634:20)
      at DeleteMutator.deleteInternalAsync (../entity/src/EntityMutator.ts:540:25)
      at ../entity/src/EntityQueryContextProvider.ts:40:22
```
1. Apply fix described above. Re-run `yarn integration` and ensure the tests now pass.

### DIFF
--- a/packages/entity-full-integration-tests/src/__integration-tests__/EntityEdgesIntegration-test.ts
+++ b/packages/entity-full-integration-tests/src/__integration-tests__/EntityEdgesIntegration-test.ts
@@ -1,108 +1,12 @@
-import {
-  EntityPrivacyPolicy,
-  ViewerContext,
-  AlwaysAllowPrivacyPolicyRule,
-  Entity,
-  EntityCompanionDefinition,
-  EntityConfiguration,
-  DatabaseAdapterFlavor,
-  CacheAdapterFlavor,
-  UUIDField,
-  EntityEdgeDeletionBehavior,
-} from '@expo/entity';
+import { ViewerContext } from '@expo/entity';
 import { RedisCacheAdapterContext } from '@expo/entity-cache-adapter-redis';
 import Redis from 'ioredis';
 import Knex from 'knex';
 import { URL } from 'url';
 
 import { createFullIntegrationTestEntityCompanionProvider } from '../testfixtures/createFullIntegrationTestEntityCompanionProvider';
-
-interface ParentFields {
-  id: string;
-}
-
-interface ChildFields {
-  id: string;
-  parent_id: string;
-}
-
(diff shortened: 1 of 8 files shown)


quinlanj (commented):
While this solves the current dependency cycle, it doesn't solve the overall problem of cycles. The solution also adds extra complexity to `inboundEdges` in order to break the current cycle. 

Since the `entity` repo is smaller than `www`, this might be a good place to try the prescribed solution in the article you linked, adding an `index/internal.ts` and have everything import from that intermediate file. This way, modules are always loaded in the same order every time.

wschurman:
> Since the `entity` repo is smaller than `www`, this might be a good place to try the prescribed solution in the article you linked, adding an `index/internal.ts` and have everything import from that intermediate file. This way, modules are always loaded in the same order every time.

The package `entity` already does this technically, but we want the library to support both use cases (repositories the both use and choose not to use `index.ts` files). Put another way, the usage of a library shouldn't dictate this convention for the using project as a whole. Therefore, we needed to find a compromise, and I think making this field a function is the simplest tradeoff.

Other solutions were considered, though:
- Entity "registry", where each entity is registered and this cyclical graph of reverse edges is built at runtime. The downside here is that the registry pattern isn't frequently used in JS and can also be hard to debug.
- Entity introspection, where a directory of concrete entity implementor definition filepaths is supplied to the framework initialization. The downside here is that this would dictate directory structure for a user's entities.

ide (approved):
This solution seems good and should hold up. There is a cyclic relationship between entity types, but it doesn't need to be resolved all at once (hence the new thunk).

Issue #90, opened 2020-11-05
Add ability to just invalidate an entity's cache


wschurman:
There will probably be times that a manual update or delete raw SQL query is needed, and the entity framework should provide a mechanism to invalidate something manually that is now known to be invalid.

PR #91, opened 2020-11-05
feat: add ability to invalidate an entity manually

wschurman:
# Why

Closes https://github.com/expo/entity/issues/90.

This may be useful when a raw SQL update is run outside the entity framework, in cases like migrations or even in application code.

# How

Add method and test.

# Test Plan

Run test.

### DIFF
--- a/packages/entity/src/EntityLoader.ts
+++ b/packages/entity/src/EntityLoader.ts
@@ -285,9 +285,18 @@ export default class EntityLoader<
   }
 
   /**
+   * Invalidate all caches for an entity. One potential use case would be to keep the entity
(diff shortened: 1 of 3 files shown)


ide (approved):
Might make sense to add another cached field to the tested entity type to ensure that both the PK and the alternative cache key field are both invalidated. But if you think that functionality is already covered by other unit tests we're probably fine.

wschurman:
> Might make sense to add another cached field to the tested entity type to ensure that both the PK and the alternative cache key field are both invalidated. But if you think that functionality is already covered by other unit tests we're probably fine.

Yep, this is tested in `EntityDataManager-test`.

PR #93, opened 2020-11-06
fix: edge deletion behavior cyclic import structure again

wschurman:
# Why

Apparently https://github.com/expo/entity/pull/89 wasn't sufficient for all cases. With the way entities are encouraged to be written, it seems that even with one side of this cycle being defined as a function isn't sufficient to squash all import cycles. I think this is a fairly unique case though as it is literally defining a cycle so I still don't think that having the whole companion or configuration be a function is necessary (don't foresee any other future features needing to specify other entities in configurations).

# How

Make the other direction edge specification a function as well.

# Test Plan

This one is harder to repro using a simple test case. To test this, I `yarn link`'ed the `entity` package into Expo code where we're seeing the behavior and ensured it fixed it.


ide (approved):
I'm a little wary of leaving the name as is instead of renaming it to `getAssociatedEntityClass` since classes are functions in JS so it's easy to get the two mixed up but this looks fine otherwise for the cyclic entity case.

quinlanj (approved):

PR #96, opened 2020-11-16
fix: fix null handling in load paths (try 2)

wschurman:
# Why

There was an issue where if during a load directive chain one of the fields was null it would call loadByIdAsync(null) which would be problematic. This opened a bit of a pandoras box of overlooked null handling issues that the typechecker couldn't handle.

First try was https://github.com/expo/entity/pull/95, but that PR made the tradeoff of breaking APIs in exchange for a cleaner implementation in `EntityAssociationLoader::loadAssociatedEntityThroughAsync`, which looking back on it was the wrong decision. Thus, this PR inverts that decision and makes the implementation slightly less pretty in exchange for this not being a breaking API change, which would've required auditing each callsite of the breakage due to the nature of the return type changes.

# How

This PR does the following:
- Add safety measures surrounding the core issue by disallowing loading entities by null values at runtime. This was only done at typecheck time in the past through the use of `NonNullable`. While this may incur a slight overhead for each load (loop through all values being loaded and ensure not null or undefined), I think the tradeoff for safety is worth it as a load by null could cause a massive amount of rows to be fetched from the database adapter.
- Add guard clauses to the implementation of some association loader functions to return early when field values for load-bys are null.

# Test Plan

Run new and old tests, typechecker.


packages/entity/src/EntityAssociationLoader.ts
```diff
@@ -54,14 +55,20 @@ export default class EntityAssociationLoader<
       .getViewerScopedEntityCompanionForClass(associatedEntityClass)
       .getQueryContextProvider()
       .getQueryContext()
-  ): Promise<Result<TAssociatedEntity>> {
+  ): Promise<
+    Result<null extends TFields[TIdentifyingField] ? TAssociatedEntity | null : TAssociatedEntity>
+  > {
     const associatedEntityID = this.entity.getField(fieldIdentifyingAssociatedEntity);
+    if (!associatedEntityID) {
+      return result(null) as any;
```
ide, line 63: Shouldn't a sufficiently knowledgeable type checker know that if `TFields[TIdentifyingField]` is nullable, then `this.entity.getField(fieldIdentifyingAssociatedEntity)` is nullable and thus returning `result(null)` is OK? If so, we might want to use `@ts-expect-error` instead of `any` here.

Is the issue that `Result<null>` doesn't satisfy `Result<TAssociatedEntity | null>`? Maybe we need some `readonly`s in Result's definition for covariance?

packages/entity/src/EntityAssociationLoader.ts
```diff
@@ -54,14 +55,20 @@ export default class EntityAssociationLoader<
...
+    }
+
     const loader = this.entity
       .getViewerContext()
       .getViewerScopedEntityCompanionForClass(associatedEntityClass)
       .getLoaderFactory()
       .forLoad(queryContext);
-    return await loader.loadByIDAsync((associatedEntityID as unknown) as TAssociatedID);
+    return (await loader.loadByIDAsync((associatedEntityID as unknown) as TAssociatedID)) as any;
```
ide, line 71: This too - I'd like to better understand why we need the `any` cast and if some simple `readonly`s can fix it.

packages/entity/src/EntityAssociationLoader.ts
wschurman: It's hard to tell how the typechecker is analyzing this, but the full message is:
```
src/EntityAssociationLoader.ts(63,7): error TS2322: Type 'Success<null>' is not assignable to type 'Result<null extends TFields[TIdentifyingField] ? TAssociatedEntity | null : TAssociatedEntity>'.
  Type 'Success<null>' is not assignable to type 'Success<null extends TFields[TIdentifyingField] ? TAssociatedEntity | null : TAssociatedEntity>'.
    Type 'null' is not assignable to type 'null extends TFields[TIdentifyingField] ? TAssociatedEntity | null : TAssociatedEntity'.
```
so it looks like it's unable to use the knowledge about `this.entity.getField(fieldIdentifyingAssociatedEntity)` when analyzing `return result(null);`.

wschurman: I'll go ahead and use `@ts-expect-error`.

packages/entity/src/EntityAssociationLoader.ts
wschurman: Temporarily changing the return type of the function to `Result<TAssociatedEntity | null>` and returning `result(null)` indicates to me that it isn't the result library (which is covariant), but instead a loss of information about the conditional type. I may be misunderstanding your suggestion though.

ide: What about `Promise<Result<TAssociatedEntity> | Result<null>>`? My question about covariance is whether `const r: Result<TAssociatedEntity | null> = result(null)` passes. I believe a correct type checker should reject this unless `Result<T>` treats `T` as `readonly`.

wschurman: Changing the return type to that and returning `r` does trigger an error.
```
src/__tests__/EntityAssociationLoader-test.ts(26,14): error TS2531: Object is possibly 'null'.
src/EntityAssociationLoader.ts(62,7): error TS2322: Type 'Success<TAssociatedEntity | null>' is not assignable to type 'Success<TAssociatedEntity> | Failure<TAssociatedEntity> | Success<null> | Failure<null>'.
  Type 'Success<TAssociatedEntity | null>' is not assignable to type 'Success<TAssociatedEntity>'.
    Type 'TAssociatedEntity | null' is not assignable to type 'TAssociatedEntity'.
      'TAssociatedEntity' could be instantiated with an arbitrary type which could be unrelated to 'TAssociatedEntity | null'.
        Type 'null' is not assignable to type 'TAssociatedEntity'.
          'TAssociatedEntity' could be instantiated with an arbitrary type which could be unrelated to 'null'.
```
This error message makes sense to me, though I need to read up on how `readonly` is used in the type system since I'm not following that piece.

ide: This part of the error makes sense to me: `Type 'Success<TAssociatedEntity | null>' is not assignable to type 'Success<TAssociatedEntity>'.` We maybe should dig into that more.

I did some testing on the side and found this checks out `const r1: Result<string | null> = result(null) as Result<null>`. I find this surprising, as it holds even if the `readonly`s are removed from `Result`:

```
    const r0 = result(1);
    const r1: Result<string | number> = r0;
    r1.value = 'hi not a number';
```

wschurman: Lol the following works:

```typescript
return result(null) as Result<
  null extends TFields[TIdentifyingField] ? TAssociatedEntity | null : TAssociatedEntity
>;
```

ide (approved):

PR #98, opened 2020-12-10
feat: transaction-scoped dataloaders

wschurman:
# Why

An interesting idea from @ide: within a transaction, subsequent reads should be able to be cached locally as long as the local cache is kept consistent (not in the cache adapter though).

# How

Keep a separate dataloader per-transaction that is invalidated synchronously.

# Test Plan

Not tested yet.

### DIFF
--- a/packages/entity/src/internal/EntityDataManager.ts
+++ b/packages/entity/src/internal/EntityDataManager.ts
@@ -4,7 +4,7 @@ import EntityDatabaseAdapter, {
   FieldEqualityCondition,
   QuerySelectionModifiers,
 } from '../EntityDatabaseAdapter';
-import { EntityQueryContext } from '../EntityQueryContext';
+import { EntityQueryContext, EntityTransactionalQueryContext } from '../EntityQueryContext';
 import EntityQueryContextProvider from '../EntityQueryContextProvider';
 import { partitionErrors } from '../entityUtils';
 import {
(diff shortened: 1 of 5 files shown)


packages/entity/src/internal/EntityDataManager.ts
```diff
@@ -192,7 +236,20 @@ export default class EntityDataManager<TFields> {
     fieldValues: readonly NonNullable<TFields[N]>[]
   ): Promise<void> {
     await this.entityCache.invalidateManyAsync(fieldName, fieldValues);
-    const dataLoader = this.getFieldDataLoaderForFieldName(fieldName);
+    const dataLoader = this.getRegularFieldDataLoderForFieldName(fieldName);
+    fieldValues.forEach((fieldValue) => dataLoader.clear(fieldValue));
+  }
+
+  private async invalidateManyByFieldEqualingTransactionalAsync<N extends keyof TFields>(
```
ide, line 243:
```suggestion
  private async invalidateManyWithinTransactionByFieldEqualingAsync<N extends keyof TFields>(
```

packages/entity/src/internal/EntityDataManager.ts
```diff
@@ -192,7 +236,20 @@ export default class EntityDataManager<TFields> {
...
+    queryContext: EntityQueryContext,
+    fieldName: N,
+    fieldValues: readonly NonNullable<TFields[N]>[]
+  ): Promise<void> {
+    if (!queryContext.isInTransaction()) {
+      return;
+    }
```
ide, line 250: Instead of returning here, it seems better not to call into the transactional methods in the first place for slightly simpler call trees. Either way there's going to be an `inTransaction()` check somewhere, so doing it earlier and higher up in the call tree seems simpler.

```ts
    // EntityMutator.ts

    if (queryContext.inTransaction()) { // <-- new
      await Promise.all([
        entityLoader.invalidateFieldsTransactionalAsync(this.originalFieldsForEntity),
        entityLoader.invalidateFieldsTransactionalAsync(this.fieldsForEntity),
      ]);
    }
```

wschurman:
Abandoning for now. While this would be a small perf increase, ~98% of loads in the EAS use case are non-transactional so it would be minimal.

wschurman:
Going to try this again. Should have significant benefits for transaction durations.

PR #101, opened 2020-12-22
feat: better database adapter error handling

wschurman:
# Why

Most ORMs parse common Postgres errors and return custom, easier-to-consume errors:
- [Sequelize](https://github.com/sequelize/sequelize/blob/febc083adee2cd3e6f24b18a556acdc4c4f50f96/src/dialects/postgres/query.js#L305)
- [Objection.js](https://github.com/Vincit/objection.js/blob/1f743815add5ac87eb6c6431f6928434ebbcb3bc/doc/recipes/error-handling.md)

This allows for the use of `instanceof`s or common entity error code handling rather than manual postgres error code handling at callsites.

For Expo in particular, we do manual error conversion of `EntityError`s in our application's top-level error handler before sending a HTTP response. With this new error handling, we can also handle database errors (both transient and permanent) in the same place rather than at each callsite (though this doesn't preclude using them in boundaries at the callsites as well for specific case application error handling logic) and we can do so by checking for these errors.

Next up will be the equivalent PR for `RedisCacheAdapter`.

# Reviewer Questions

- [x] Should the core framework (`entity`) have base classes for these specific Postgres errors? For example, should there be a base class `EntityDatabaseAdapterNotNullConstraintError` that `PostgresEntityDatabaseAdapterNotNullConstraintError` extends? The reason I see for doing so is other concrete database adapter implementations could throw the same base class, and consumers could `instanceof` just the base class for that whole class of error. The reason I didn't do it is we currently only have one concrete database adapter (postgres) so it's hard to predict good base classes between theoretical database adapters.
- [ ] What is the best way to wrap and re-throw an error? I transferred the stack trace and message (via the constructor). Are there other things that should be transferred?

# How

Add new error types, add wrapper function to catch and re-throw errors, add integration tests for all error types.

# Test Plan

Run integration test.


packages/entity/src/errors/EntityError.ts
```diff
@@ -0,0 +1,11 @@
+import ES6Error from 'es6-error';
+
+export enum EntityErrorState {
+  UNKNOWN,
+  TRANSIENT,
```
ide, line 5: In some systems, retryable errors come with information about when to retry (this is most common with 429s). Something to think about.

ide (commented):
In most JS environments it's not common to use `instanceof` with errors the way one might with Java, for example. It's more common in the browser APIs and Node APIs to attach error codes to Error objects. Node's pattern is to define a field named "code" that has a value like "ERR_TOPIC_CAUSE" (there are some older error values like "ECONNRESET" that don't follow this pattern especially if they are based on the codes from the OS layer below).

wschurman:
> In most JS environments it's not common to use `instanceof` with errors the way one might with Java, for example. It's more common in the browser APIs and Node APIs to attach error codes to Error objects. Node's pattern is to define a field named "code" that has a value like "ERR_TOPIC_CAUSE" (there are some older error values like "ECONNRESET" that don't follow this pattern especially if they are based on the codes from the OS layer below).

Yeah, I've definitely seen error codes more frequently in system code, though it does seem that third-party JS ORMs are suggesting using `instanceof` more commonly, at least in the more modern ones. The two drawbacks that I can think of when using codes are:
- Less type safety. The type would have to look something like `type ErrBlah = Error & { code: 'BLAH'};` and then use discrimination to infer the final type from an ambiguous type. With instanceof, the type safety is a bit clearer.
- Less ability to do error hierarchies. For example, in application code we could do an `instanceof EntityError` to catch all errors belonging to the entity framework (as we do today), `instanceof EntityDatabaseAdapterError` to catch just entity errors coming from the database adapter, or we can catch specific database adapter errors. With error codes, this becomes a bit more difficult as multiple codes are needed on each error (`isEntityError`, `groupCode`, `errorCode`, ...), and the hierarchical structure is less flexible.

packages/entity/src/errors/EntityError.ts
wschurman: Yeah, not sure we have quite enough information at the entity level to determine what a good retry timeout would be. I think this would probably be better determined at the application layer (since it does the Knex configuration) and then also added to the 429s at that layer.

packages/entity-database-adapter-knex/src/errors/PostgresEntityDatabaseAdapterError.ts
```diff
@@ -0,0 +1,36 @@
+import { EntityDatabaseAdapterError, EntityErrorState } from '@expo/entity';
+
+export default abstract class PostgresEntityDatabaseAdapterError extends EntityDatabaseAdapterError {
+  constructor(public readonly nativeError: Error) {
```
ide, line 4: An early-ish Stage 2 proposal uses `cause`: https://github.com/tc39/proposal-error-cause

ide (changes requested):
Requesting changes to add `code` to line up with Node, but not necessarily removing the classes.

The code is useful in that it's printed out with console.log and is more in line with Node's new convention: https://nodejs.org/api/errors.html#errors_node_js_error_codes. The code isn't mutually exclusive from class hierarchies so we could combine the two, but I also think that classes don't offer that much and are less flexible than structural types. In general, `type` in TS is more flexible and since TS 3.x or so, has almost all of the capabilities of `interface` so there's almost reason to use the latter except for with `implements`.

I do think, with or without classes, it would be useful to define codes like `ERR_ENTITY_DATABASE_UNIQUE_CONSTRAINT` so that this library lines up better with Node's conventions.

In our API server, I also would like to avoid deep error hierarchies, namely avoid Java's exception system, and instead opt to identify a few broad classes of errors and cross-cutting fields. Example classes: user's fault, infra outage, programmer error. Example fields: error code, retryability, error causes (useful for when we want to see why one error caused another), HTTP status code, whether the error is user-facing. Errors won't have all of these fields, but I'd expect many of them to have them.

ide (approved):

PR #102, opened 2020-12-23
feat: better cache adapter error handling

wschurman:
# Why

The cache adapter equivalent to https://github.com/expo/entity/pull/101. Redis/ioredis doesn't throw any specific errors with codes, just a single error type `ReplyError` with a message (https://github.com/luin/ioredis#error-handling) so we just assume that all redis errors are transient.

# How

Add new error class and error code, wrap ioredis calls, add test.

# Test Plan

Run test.


packages/entity-cache-adapter-redis/src/errors/wrapNativeRedisCall.ts
```diff
@@ -0,0 +1,11 @@
+import { EntityCacheAdapterTransientError } from '@expo/entity';
+
+export default async function wrapNativeRedisCall<T>(fn: Promise<T>): Promise<T> {
```
ide, line 3:
```suggestion
export default async function wrapNativeRedisCall<T>(promise: Promise<T>): Promise<T> {
```

This said, a slightly more robust pattern would be to write

```ts
export default async function wrapNativeRedisCall<T>(fn: () => Promise<T>): Promise<T> {

wrapNativeRedisCall(() => this.context.redisClient.mget(...redisKeys));
```

This handles the case where the promise-creating function throws before it first yields. This matters for non-async functions like:
```ts
function fn(): Promise {
  assert(something);
  return new Promise(...);
}

fn(); // throws before we even get back a promise
```

ide (approved):

PR #108, opened 2021-01-20
fix: move runInTransaction to ViewerContext

wschurman:
# Why

This has been suggested by @jkhales and @ide a few times as a better pattern. At its core this does the same thing as the previous implementation, it just removes the unnecessary association with concrete entity types.

# How

Move the methods to `ViewerContext`. Note that this will most likely be used in methods in `ViewerContext` subclasses in the following manner:

```typescript
// in ExpoViewerContext.ts

async runInPostgresTransactionAsync<TResult>(
  transactionScope: (queryContext: EntityTransactionalQueryContext) => Promise<TResult>
): Promise<TResult> {
  return await this.runInTransactionForDatabaseAdaptorFlavorAsync(
    DatabaseAdapterFlavor.POSTGRES,
    transactionScope
  );
}
```

so that callsites don't need to specify the `DatabaseAdapterFlavor`. 

# Questions for Reviewers

Do you think it's worth added these `DatabaseAdapterFlavor`-hardcoded methods to `ViewerContext`? Since the `DatabaseAdapterFlavor` is hardcoded in this package, it can be done cleanly. The upside is that consumers don't need to specify them. The downside is that the number of methods would grow if we add more `DatabaseAdapterFlavor`s.

# Test Plan

Run all tests.

### DIFF
--- a/packages/entity/src/EntityCompanionProvider.ts
+++ b/packages/entity/src/EntityCompanionProvider.ts
@@ -226,6 +226,13 @@ export default class EntityCompanionProvider {
     });
   }
 
+  getQueryContextProviderForDatabaseAdaptorFlavor(
+    databaseAdapterFlavor: DatabaseAdapterFlavor
+  ): EntityQueryContextProvider {
+    const entityDatabaseAdapterFlavor = this.databaseAdapterFlavors[databaseAdapterFlavor];
+    return entityDatabaseAdapterFlavor.queryContextProvider;
+  }
+
   private getTableDataCoordinatorForEntity<TFields>(
     entityConfiguration: EntityConfiguration<TFields>,
     entityClassName: string
--- a/packages/entity-database-adapter-knex/src/__integration-tests__/PostgresEntityIntegration-test.ts
+++ b/packages/entity-database-adapter-knex/src/__integration-tests__/PostgresEntityIntegration-test.ts
@@ -3,6 +3,7 @@ import {
   createUnitTestEntityCompanionProvider,
   enforceResultsAsync,
   ViewerContext,
+  DatabaseAdapterFlavor,
 } from '@expo/entity';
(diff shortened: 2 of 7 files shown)


jkhales (approved):
nice. ✨ 

I would be curious to know what other flavors will be added. For example if MySQL is added then things would be similar enough for transactions that a `runInSQLTransactionAsync` method seems like the cleaner abstraction.

wschurman:
> I would be curious to know what other flavors will be added. For example if MySQL is added then things would be similar enough for transactions that a `runInSQLTransactionAsync` method seems like the cleaner abstraction.

We'd need one method per flavor just in case an application uses multiple flavors. This whole flavor concept as a whole though probably needs to be removed and just replaced with strings so that client applications can have as many "flavors" as they want and configure them however they like.

packages/entity/src/EntityCompanionProvider.ts
```diff
@@ -226,6 +226,13 @@ export default class EntityCompanionProvider {
     });
   }
 
+  getQueryContextProviderForDatabaseAdaptorFlavor(
```
ide, line 229: Might want to make `getTableDataCoordinatorForEntity` call this so that the behavior between the two methods stays in sync.

ide:
Since the number of database flavors will likely stay small, having hardcoded methods seems good so that we don't need to pass the same parameter in everywhere. It also might help with type hints a bit if the Postgres query context were to be different than a MySQL query context.

Longer-term, if we have multiple flavors, I think we will want to automatically detect what flavors are used and just have a generic "runInTransactionAsync" method that creates the correct query contexts on the fly as needed, or at least warn if you started a Postgres transaction but tried to access MySQL. There might be cases where we need to let developers opt out (e.g. they want a PG transaction but not a MySQL one) but those cases seem rare.

PR #113, opened 2021-01-29
feat: add basic field type runtime validators

wschurman:
# Why

We're seeing a decent number of bugs/crashes surrounding invalid types of values being set in mutators and invalid types of load-by values being used in loaders.

These bugs manifest themselves in obvious ways sometimes, and other times less obvious. 
- An example on read might look like: `select * from "table" where "id" = ANY($1) - invalid input syntax for uuid: "c6cef326-2a93-4477-8e46-7faae8e1684"`
- An example on write looks like: `TypeError: Cannot read property 'push' of undefined`

The write example is a fairly clear error already, but this PR preempts the round trip to the DB by detecting a query that will most likely fail anyways ahead of time.

For the read example, the error is more cryptic and is what triggered the investigation and this PR as the result. The issue is that when a load comes in for a UUID that postgres thinks is valid, but postgres returns the UUID in the normalized format in the result set, we don't know to create a map entry at https://github.com/expo/entity/blob/master/packages/entity/src/EntityDatabaseAdapter.ts#L123 so the `push` call a few lines later will fail. Concretely, the trace is:
- Request comes in for UUID `a0eebc999c0b4ef8bb6d6bb9bd380a11` which postgres thinks is valid.
- The code in entity creates a map entry for `a0eebc999c0b4ef8bb6d6bb9bd380a11`.
- Postgres returns row with normalized UUID `a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11`.
- We look in the map expecting the array to exist but it doesn't since they're different strings.

Note that this postgres behavior is documented at https://www.postgresql.org/docs/9.1/datatype-uuid.html

# How

Add field value validation to the following calls:
- `EntityMutator.createAsync/updateAsync`
- `EntityLoader.loadByFieldValueAsync` and similar methods.

# Test Plan

Run tests, including new ones to test behavior and validation correctness.


packages/entity/src/EntityMutator.ts
```diff
@@ -72,6 +74,19 @@ abstract class BaseMutator<
...
   ) {}
 
+  protected validateFields(fields: Partial<TFields>): Result<void> {
+    for (const fieldName in fields) {
+      const fieldValue = fields[fieldName];
+      const fieldDefinition = this.entityConfiguration.schema.get(fieldName);
+      invariant(fieldDefinition, `must have field definition for field = ${fieldName}`);
+      const isInputValid = fieldDefinition.validateInputValueIfNotNull(fieldValue);
+      if (!isInputValid) {
+        return result(new EntityInvalidFieldValueError(this.entityClass, fieldName, fieldValue));
```
ide, line 84: May want to consider using AggregateError (https://www.npmjs.com/package/aggregate-error, also will be built into the next version of Node) if we want to return one error for multiple fields. But we'd need to make our error-handling code know how to figure out things like transient vs. permanent AggregateErrors.

packages/entity/src/EntityLoader.ts
```diff
@@ -305,4 +355,32 @@ export default class EntityLoader<
       }
     });
   }
+
+  private partitionFieldValues<N extends keyof Pick<TFields, TSelectedFields>>(
```
ide, line 359: suggestion:
```suggestion
  private partitionFieldValuesByValidity<N extends keyof Pick<TFields, TSelectedFields>>(
```

packages/entity/src/EntityLoader.ts
```diff
@@ -199,6 +224,31 @@ export default class EntityLoader<
     fieldEqualityOperands: FieldEqualityCondition<TFields, N>[],
     querySelectionModifiers: QuerySelectionModifiers<TFields> = {}
   ): Promise<readonly Result<TEntity>[]> {
+    const invalidFieldValueErrors: Result<TEntity>[] = [];
```
ide, line 227: I'm not sure we want to return an array of failure results since each error doesn't correspond to an entity that failed to load, it corresponds to an invalid query field. I think of `Result<TEntity>` as containing an entity or the error saying why we couldn't load that specific entity.

It doesn't make sense to iterate over `invalidFieldValueErrors` the same way it'd make sense to iterate over an array of entities and failures to load those entities. Perhaps we should just construct an AggregateError here and throw, or throw an unaggregated error on the first invalid value we see.

quinlanj (commented):
Agree with @ide about the AggregateError. 

It'd make sense to throw an `InvalidFieldValueError` if `www` is not able to fix or detect the error without making a trip to Postgres. An example of this is corrupt data in postgres itself. 

But in this case, it should be the responsibility of `www` to make sure the field inputs are correct. The entity framework should throw an error without wrapping it in a result, and only request data from the DB when it knows all the inputs are well formed.

We also discussed this on zoom, and found out dataloaders has this behaviour, so this will be a happy case of accidental consistency lol

wschurman:
@quinlanj and I talked a bit about this and think that throwing independently of Result actually makes the most sense for two reasons:
- This is in some sense programmer error (passing an invalid value into the framework). The framework returns `Result<Error>` only for things that are valid framework calls but yield either not found or not authorized. For things like database misconfiguration (https://github.com/expo/entity/blob/5208aeedb0192f0460ba3d575770a04175f5c3aa/packages/entity/src/EntityLoader.ts#L139) we throw an error instead.
- The current behavior is actually to throw an error for the cases specified in the summary.

I'll update the PR to reflect this.

ide (approved):

quinlanj (approved):

PR #125, opened 2021-05-26
fix: upgrade ioredis, knex, and pg

wschurman:
# Why

It's good to be on recent versions of libraries for perf and security reasons. This PR updates three dependencies:
- `ioredis` (& types) - 4.17.3 -> 4.27.3 - All bug fixes: https://github.com/luin/ioredis/blob/master/Changelog.md
- `pg` - 8.3.2 -> 8.6.0 - Small bug fixes: https://github.com/brianc/node-postgres/blob/master/CHANGELOG.md
- `knex` - 0.21.5 -> 0.95.6 - Major upgrade
    - Changelog: https://github.com/knex/knex/blob/master/CHANGELOG.md#0950---03-march-2021
    - Upgrade guide: https://github.com/knex/knex/blob/master/UPGRADING.md
    - Mostly default -> named export changes for typescript and `knex.raw`
    - The following should be fine:
        > Trying to execute an operation resulting in an empty query such as inserting an empty array, will now throw an error on all database drivers.
    

# How

Upgrade packages, `yarn tsc`, run tests.

# Test Plan

`yarn tsc`, run tests. `yarn link` into Expo www, run all tests there as well.


ide (approved):
I don't totally understand the decisions the type declaration authors made for knex vs Knex (I get why they are separate but not why the two are both namespaces) but this commit looks fine.

packages/entity-database-adapter-knex/src/__integration-tests__/PostgresEntityIntegration-test.ts
```diff
@@ -5,7 +5,7 @@ import {
   ViewerContext,
 } from '@expo/entity';
 import { enforceAsyncResult } from '@expo/results';
-import Knex from 'knex';
+import { knex, Knex } from 'knex';
```
jkhales, line 8: wat! 😅

PR #130, opened 2021-06-28
feat: allow null field values in loadManyByFieldEqualityConjunctionAsync

wschurman:
# Why

A feature request from @jkhales for an API endpoint within the Expo server. 

Since the results for this type of load aren't stored in a dataloader or cached, we can just pipe null fieldValues directly through to the database adapters for load, and it is their duty to handle them for this type of load.

# How

This ended up being slightly more difficult than anticipated but not too bad. Complications:
- `ANY(?)` cannot contain `NULL` so it needs to be split into separate `OR WHERE` but within the same conjunctive clause.
- Replicating the default `NULLS FIRST/LAST` postgres behavior in StubDatabaseAdapter required checking for null explicitly: https://www.postgresql.org/docs/8.3/queries-order.html#:~:text=The-,NULLS%20FIRST,-and%20NULLS%20LAST

# Test Plan

Run tests.

### DIFF
--- a/packages/entity-database-adapter-knex/src/PostgresEntityDatabaseAdapter.ts
+++ b/packages/entity-database-adapter-knex/src/PostgresEntityDatabaseAdapter.ts
@@ -89,10 +89,24 @@ export default class PostgresEntityDatabaseAdapter<TFields> extends EntityDataba
 
     if (tableFieldSingleValueEqualityOperands.length > 0) {
       const whereObject: { [key: string]: any } = {};
-      for (const { tableField, tableValue } of tableFieldSingleValueEqualityOperands) {
-        whereObject[tableField] = tableValue;
+      const nonNullTableFieldSingleValueEqualityOperands = tableFieldSingleValueEqualityOperands.filter(
+        ({ tableValue }) => tableValue !== null
+      );
+      const nullTableFieldSingleValueEqualityOperands = tableFieldSingleValueEqualityOperands.filter(
+        ({ tableValue }) => tableValue === null
+      );
+
+      if (nonNullTableFieldSingleValueEqualityOperands.length > 0) {
+        for (const { tableField, tableValue } of nonNullTableFieldSingleValueEqualityOperands) {
+          whereObject[tableField] = tableValue;
+        }
+        query = query.where(whereObject);
       }
-      query = query.where(whereObject);
+      if (nullTableFieldSingleValueEqualityOperands.length > 0) {
+        for (const { tableField } of nullTableFieldSingleValueEqualityOperands) {
+          query = query.whereNull(tableField);
+        }
+      }
     }
 
     if (tableFieldMultiValueEqualityOperands.length > 0) {
@@ -102,7 +116,7 @@ export default class PostgresEntityDatabaseAdapter<TFields> extends EntityDataba
           builder.whereRaw('?? = ANY(?)', [tableField, [...nonNullTableValues]]);
           // there was at least one null, allow null in this equality clause
           if (nonNullTableValues.length !== tableValues.length) {
-            builder.orWhere(tableField, null);
+            builder.orWhereNull(tableField);
           }
         });
       }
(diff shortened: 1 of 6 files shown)


packages/entity/src/utils/testing/__tests__/StubDatabaseAdapter-test.ts
```diff
@@ -192,6 +192,81 @@ describe(StubDatabaseAdapter, () => {
...
+                  testIndexedField: 'h3',
+                  numberField: 3,
+                  stringField: 'a',
+                  dateField: new Date(),
+                  nullableField: null,
+                },
+                {
+                  customIdField: '3',
+                  testIndexedField: 'h3',
+                  numberField: 3,
```
ide, line 233: Make this row distinguishable from the row above:
```suggestion
                  customIdField: '4',
                  testIndexedField: 'h4',
                  numberField: 4,
```

packages/entity/src/utils/testing/StubDatabaseAdapter.ts
```diff
@@ -76,13 +76,15 @@ export default class StubDatabaseAdapter<T> extends EntityDatabaseAdapter<T> {
     const bField = objectB[currentOrderBy.columnName];
     switch (currentOrderBy.order) {
       case OrderByOrdering.DESCENDING:
-        return aField > bField
+        // simulate NULLS FIRST for DESC
+        return aField > bField || aField === null
```
ide, line 80: null is ordered somewhere between -1 and 0 when comparing it against integers. I think it is clearer to do the null checks first (if both are null, return 0, if either one is null, return 1 or -1) and then run the regular comparison checks once we know both values are non-null.

As it stands I think this current implementation breaks down with aField = 1, bField = null.

(worth adding a regression test perhaps)

ide (changes requested):

jkhales (approved):
Wow, quick turn around! Thank you!

packages/entity-database-adapter-knex/src/PostgresEntityDatabaseAdapter.ts
```diff
@@ -97,7 +97,14 @@ export default class PostgresEntityDatabaseAdapter<TFields> extends EntityDataba
 
     if (tableFieldMultiValueEqualityOperands.length > 0) {
       for (const { tableField, tableValues } of tableFieldMultiValueEqualityOperands) {
-        query = query.whereRaw('?? = ANY(?)', [tableField, [...tableValues]]);
+        const nonNullTableValues = tableValues.filter((tableValue) => tableValue !== null);
+        query = query.where((builder) => {
+          builder.whereRaw('?? = ANY(?)', [tableField, [...nonNullTableValues]]);
+          // there was at least one null, allow null in this equality clause
+          if (nonNullTableValues.length !== tableValues.length) {
+            builder.orWhere(tableField, null);
```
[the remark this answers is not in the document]
ide: Good catch. I agree IS NULL's explicitness is appealing and what I'm used to writing. Maybe knex/pg handles `orWhere(x, null)` as a special case but using the dedicated `orWhereNull` function matches our intent.

packages/entity/src/utils/testing/StubDatabaseAdapter.ts
ide: At this point we are not guaranteed `aField !== null`:
```suggestion
        return aField > bField
```

packages/entity/src/utils/testing/StubDatabaseAdapter.ts
```diff
@@ -75,18 +75,38 @@ export default class StubDatabaseAdapter<T> extends EntityDatabaseAdapter<T> {
...
+        // simulate NULLS LAST for ASC
+        if (aField === null && bField === null) {
+          return 0;
+        } else if (bField === null) {
+          return -1;
+        } else if (aField === null) {
+          return 1;
+        }
+
+        return bField > aField || bField === null
```
ide, line 104:
```suggestion
        return bField > aField
```

ide (approved):

PR #134, opened 2021-09-07
feat: upgrade TypeScript to 4.4

wschurman:
# Why

We're a few versions behind. This comes with quite a few new features, but the ones that affect this repo or are likely to affect it are:
- [`noPropertyAccessFromIndexSignature`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-2.html#--nopropertyaccessfromindexsignature): stricter rules around property access on not-well-specified object types. Helps with property misspellings mostly.
- [`noUncheckedIndexedAccess`](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess): stricter index access checks where array indexing can return undefined. This is much more in alignment with this codebase's style than unchecked, but I can see how it may not be appropriate for all codebases.
- [`exactOptionalPropertyTypes`](https://www.typescriptlang.org/tsconfig#exactOptionalPropertyTypes): Makes assigning undefined to an object property value need to specify it as a possible RHS value rather than the odd left-hand `?` operator. This is more along the lines of how I think about typesystems working with JS objects, and is how this codebase was written in most cases (fixes a couple). This helps with a rare but potential bug due to a difference between a property existing and being assigned to undefined and not existing.
- Always truthy promise checks https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-3.html#always-truthy-promise-checks - We already have a lint rule for this I believe, but this codifies it in the typechecker which is better.
- `override` (next PR): https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-3.html#override-and-the---noimplicitoverride-flag
- defaulting to `useUnknownInCatchVariables` in strict: Since JS can throw non-errors this is more correct, and even if we don't expect/allow first-party code to throw non-errors, libraries may still do it. This helps us catch odd bugs where that may be the case.

In addition to those, it is faster in general.

# How

Upgrade the package. Ensure all tests and docs generation pass.

# Test Plan

Wait for CI.

### DIFF
--- a/packages/entity-database-adapter-knex/src/__integration-tests__/PostgresInvalidSetup-test.ts
+++ b/packages/entity-database-adapter-knex/src/__integration-tests__/PostgresInvalidSetup-test.ts
@@ -1,6 +1,7 @@
 import { ViewerContext } from '@expo/entity';
 import { enforceAsyncResult } from '@expo/results';
 import { knex, Knex } from 'knex';
+import nullthrows from 'nullthrows';
 
 import InvalidTestEntity from '../testfixtures/InvalidTestEntity';
 import { createKnexIntegrationTestEntityCompanionProvider } from '../testfixtures/createKnexIntegrationTestEntityCompanionProvider';
@@ -12,11 +13,11 @@ describe('postgres entity integration', () => {
     knexInstance = knex({
       client: 'pg',
       connection: {
-        user: process.env.PGUSER,
-        password: process.env.PGPASSWORD,
+        user: nullthrows(process.env['PGUSER']),
+        password: nullthrows(process.env['PGPASSWORD']),
         host: 'localhost',
-        port: parseInt(process.env.PGPORT!, 10),
-        database: process.env.PGDATABASE,
+        port: parseInt(process.env['PGPORT']!, 10),
+        database: nullthrows(process.env['PGDATABASE']),
       },
     });
   });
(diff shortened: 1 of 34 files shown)


packages/entity-database-adapter-knex/src/__integration-tests__/PostgresInvalidSetup-test.ts
```diff
@@ -12,11 +13,11 @@ describe('postgres entity integration', () => {
...
       client: 'pg',
       connection: {
-        user: process.env.PGUSER,
-        password: process.env.PGPASSWORD,
+        user: nullthrows(process.env['PGUSER']),
+        password: nullthrows(process.env['PGPASSWORD']),
         host: 'localhost',
-        port: parseInt(process.env.PGPORT!, 10),
-        database: process.env.PGDATABASE,
+        port: parseInt(process.env['PGPORT']!, 10),
```
ide, line 19:
```suggestion
        port: parseInt(nullthrows(process.env['PGPORT']), 10),
```

packages/entity/src/EntityDatabaseAdapter.ts
```diff
@@ -69,12 +69,14 @@ export interface QuerySelectionModifiers<TFields> {
...
-  offset?: number;
-  limit?: number;
+  orderBy:
+    | {
+        columnName: string;
+        order: OrderByOrdering;
+      }[]
+    | undefined;
+  offset: number | undefined;
+  limit: number | undefined;
```
ide, line 79: For APIs it's usually nicer to be able to omit a field than to need to write `{ unusedOption: undefined }`. Depending on how this code is called into, it might be better to write: `{ orderBy?: {...} | undefined }` to allow both.

ide (approved):

jkhales:
🥳

PR #135, opened 2021-09-07
feat: enable noImplicitOverride tsc setting

wschurman:
# Why

This enables noImplicitOverride as described in https://github.com/expo/entity/pull/134.

# How

Enable the flag, fix tsc.

# Test Plan

Wait for CI tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "tsc": "lerna run tsc",
     "clean": "lerna run clean",
     "lint": "lerna run lint",
+    "lint-fix": "lerna run lint-fix",
     "test": "lerna run test",
     "integration": "./resources/run-with-docker lerna run integration-no-setup",
     "prepublish": "lerna link && lerna run clean",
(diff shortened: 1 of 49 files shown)


ide (commented):
If a type declaration (d.ts) includes `override`, can older version of TypeScript use those declarations? Asking since this change might require upgrading www's version of TS in order to use these changes.

Looks like Prettier needs to run over all the files again.

ide (approved):

wschurman:
> If a type declaration (d.ts) includes `override`, can older version of TypeScript use those declarations? Asking since this change might require upgrading www's version of TS in order to use these changes.

The way I understand it this is the inverse strategy of some other languages `open` (kotlin), `virtual` (C#), etc since it is only declared on the subclass and is only checked against the superclass if the subclasses tsconfig includes the directive. What this means for libraries is that unless the destination project has the directive and the member has the modifier it won't check that it overrides the library's member.

As for whether the override modifier is included in the .d.ts for completeness even though it doesn't have an effect on consumers, I don't think it does, but again it shouldn't matter since it only affects the typechecking of that class against its parent.

PR #136, opened 2021-09-17
feat: use better QueryContext types for transactional triggers and validators

wschurman:
# Why

There is a use case for triggers to run in two phases:
1. Phase 1 runs during the transaction, and collects data for use in phase 2
2. Phase 2 runs after the transaction and operates on data collected in phase 1, but in a non-transactional context

The way to implement this is with a transactional trigger that calls `queryContext.appendPostTransactionCallback` for phase 2. That method is only available on `EntityTransactionalQueryContext`, and since we can be guaranteed that these types of triggers have a transactional query context we can type it as such.

# How

Update the types to match reality.

# Test Plan

`yarn tsc`


ide (approved):
LGTM. I also could see a hypothetical use case for running code in cases without query contexts as well, since every Postgres query makes a transaction if it's not already in one.

PR #138, opened 2021-10-08
feat: add IP support

kgc00:
# Why

Planning to land a series of PRs in [universe](https://github.com/expo/universe/pull/8245) around storing IPs in PG.

# How

Added the new type.

# Test Plan

Added tests to confirm the validation works as expected.

### DIFF
--- a/packages/entity-ip-address-field/src/__tests__/EntityFields-test.ts
+++ b/packages/entity-ip-address-field/src/__tests__/EntityFields-test.ts
@@ -1,4 +1,4 @@
-import describeFieldTestCase from '@expo/entity/build/utils/testing/describeFieldTestCase';
+import { describeFieldTestCase } from '@expo/entity';
 
 import { IPAddressField } from '../EntityFields';
 


kgc00:
> I'm tempted to either put this in www for now or put it in its own package for two reasons:
> 
> 1. `is-cidr` is a fairly specialized dependency that isn't core to entity functionality.
> 2. This seems like it's in early stages and could be prone to change a bit in the future (allow for v6 addresses for example). Until the API is settled, it may be easier to develop independently of the @expo/entity package so that iteration is easier (releasing this package is a pain).
> 
> My gut says start with this in www and then once ironed out move to its own package in this repo.

haha, you got to the review a minute too early! So the updates: 
- i just changed the type check to keep it in parity with how PG acts when you try to add an `INET` value. This means no more custom 'dont allow this kind of address (ipv6)'. 
- I also updated the dependency from `is-cidr` to `is-ip` as it's a better fit for the job and helps us stay in parity with PG behavior.

Happy to leave this here as an open PR until I'm done with he other stuff in universe though!

PR #144, opened 2021-12-22
fix: separate post-commit invalidation callbacks from trigger and user callbacks

wschurman:
# Why

The issue we're seeing is:
- A post-commit trigger is calling load on data that was deleted. This call should be possible (using loadNullable).
- The cache still has data in it since the ordering of post-commit callbacks is undefined.
- The data is loaded even though it no longer exists.

# How

To fix this, add some strict ordering to post-commit callbacks to ensure that post-commit invalidations run before any other callbacks. This ensures that the post-commit callbacks that access the cache have consistent cache loads.

# Test Plan

- Run the new test without the fix, see that it fails (the exception is thrown)
- Run the new test with the fix, see that it passes.

### DIFF
--- a/packages/entity/src/__tests__/EntityMutator-MutationCacheConsistency-test.ts
+++ b/packages/entity/src/__tests__/EntityMutator-MutationCacheConsistency-test.ts
@@ -1,0 +1,105 @@
+import Entity from '../Entity';
+import { EntityCompanionDefinition } from '../EntityCompanionProvider';
+import EntityConfiguration from '../EntityConfiguration';
+import { UUIDField } from '../EntityFields';
+import { EntityNonTransactionalMutationTrigger } from '../EntityMutationTriggerConfiguration';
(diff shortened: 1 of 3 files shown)


packages/entity/src/__tests__/EntityMutator-MutationCacheConsistency-test.ts
```diff
@@ -0,0 +1,105 @@
...
+        throw new Error(
+          'should not have been able to re-load the entity after delete. this means the cache has not been cleared'
+        );
+      }
+    }
+  }
+}
+
+describe('EntityMutator', () => {
+  test('cache concistency with post-commit callbacks', async () => {
```
ide, line 93:
```suggestion
  test('cache consistency with post-commit callbacks', async () => {
```

ide (approved):
Sounds right. Triggers logically don't care about the cache. There still is a race condition where another process reading the same data at this time could fill the cache, but we have that problem anyway.

PR #145, opened 2021-12-23
feat: add cascading deletion info

wschurman:
# Why

The use case for this is as follows:
- We have a trigger that we want to execute to prevent deletion of an entity in certain cases but not others.
- In a normal deletion case, we want to prevent deletion if it is the last remaining entity of its type.
- In a cascade deletion case, we want to allow deletion even if it is the last remaining entity of its type so that it doesn't become orphaned.

# How

This feature lets us solve this by inspecting the value of the cascade to see what case it's being deleted in. If it is being deleted in the normal case (no cascading info or different type of cascade) then we block the deletion in a beforeDelete trigger. If it is being deleted in a cascade deletion case we allow the deletion.

# Test Plan

Run tests.

### DIFF
--- a/packages/entity-database-adapter-knex/src/testfixtures/PostgresTriggerTestEntity.ts
+++ b/packages/entity-database-adapter-knex/src/testfixtures/PostgresTriggerTestEntity.ts
@@ -10,7 +10,7 @@ import {
   EntityMutationTrigger,
   EntityQueryContext,
   EntityNonTransactionalMutationTrigger,
-  EntityMutationInfo,
+  EntityTriggerMutationInfo,
 } from '@expo/entity';
 import { Knex } from 'knex';
 
@@ -111,7 +111,7 @@ class ThrowConditionallyTrigger extends EntityMutationTrigger<
     _viewerContext: ViewerContext,
     _queryContext: EntityQueryContext,
     entity: PostgresTriggerTestEntity,
-    _mutationInfo: EntityMutationInfo<
+    _mutationInfo: EntityTriggerMutationInfo<
       PostgresTriggerTestEntityFields,
       string,
       ViewerContext,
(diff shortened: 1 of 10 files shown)


ide (approved):
Code looks OK.

One possible gotcha that comes to mind is that we often express `ON DELETE CASCADE` in Postgres and use `CASCADE_DELETE_INVALIDATE_CACHE` in Entities, the logic being: let Postgres's schema be the source of truth of how to handle foreign-key deletions, and invalidate the entity cache to let the next read operation fill it with the correct value.

`CASCADE_DELETE_INVALIDATE_CACHE` doesn't actually issue any `DELETE` queries. It might not delete any entities at all (except for the first one being deleted that started the cascade).

So, if triggers want to behave differently based on whether a row is actually being deleted, the entity's inbound FK edges needs to be annotated with `CASCADE_DELETE` or the programmer needs to be sure that Postgres has an `ON DELETE CASCADE` constraint.

(The motivation for this PR sounds like it is to distinguish between cascading and non-cascading operations, which shouldn't be affected by the behavior described above, which is just something else to watch out for.)

wschurman:
> Code looks OK.
> 
> One possible gotcha that comes to mind is that we often express `ON DELETE CASCADE` in Postgres and use `CASCADE_DELETE_INVALIDATE_CACHE` in Entities, the logic being: let Postgres's schema be the source of truth of how to handle foreign-key deletions, and invalidate the entity cache to let the next read operation fill it with the correct value.
> 
> `CASCADE_DELETE_INVALIDATE_CACHE` doesn't actually issue any `DELETE` queries. It might not delete any entities at all (except for the first one being deleted that started the cascade).
> 
> So, if triggers want to behave differently based on whether a row is actually being deleted, the entity's inbound FK edges needs to be annotated with `CASCADE_DELETE` or the programmer needs to be sure that Postgres has an `ON DELETE CASCADE` constraint.
> 
> (The motivation for this PR sounds like it is to distinguish between cascading and non-cascading operations, which shouldn't be affected by the behavior described above, which is just something else to watch out for.)

Yeah, I think for now we assume that `CASCADE_DELETE_INVALIDATE_CACHE` will always have a corresponding `ON DELETE CASCADE` constraint. I think in an ideal world we wouldn't even have `CASCADE_DELETE_INVALIDATE_CACHE` and this wouldn't be an issue.

As for this PR, we should continue with this assumption and assume that all cascades detected by entity will actually match those done in the database.

PR #147, opened 2021-12-30
feat: add pre-commit callbacks on EntityQueryContext

wschurman:
# Why

There are a lot of pieces of logic in our code that are similar in nature and contain all the following qualities:
- Execute during a postgres transaction (transactional query context)
- Non-transactional call to third-party service/dependency that throws when call is invalid or service is down. Think a call to stripe, bigquery log, etc.
- Critical for consistency. If the third party call throws, the transaction should roll back.
- Should be run as late as possible in the transaction. Or, put another way, if the transaction fails (code in transaction block throws) prior to this, this code should not be run.
- There are places where we have multiple of these types of calls and they need to be ordered according to some heuristic that the application determines.

Triggers unfortunately don't have the ability to specify when they should be run in relation to other triggers or during cascades (by design, I think, since it is tough to schedule cascading triggers). There are post-commit triggers/callbacks but those run outside the transaction.

# How

Introduce a new concept, a "Pre-commit callback". These will run serially in the order specified by the application right before the transaction commits. If one throws, subsequent ones won't be run and the transaction will be rolled-back.

# Test Plan

Run new tests.

### DIFF
--- a/packages/entity/src/EntityQueryContext.ts
+++ b/packages/entity/src/EntityQueryContext.ts
@@ -1,6 +1,10 @@
 import EntityQueryContextProvider from './EntityQueryContextProvider';
 
 export type PostCommitCallback = (...args: any) => Promise<any>;
+export type PreCommitCallback = (
+  queryContext: EntityTransactionalQueryContext,
+  ...args: any
+) => Promise<any>;
 
 /**
  * Entity framework representation of transactional and non-transactional database
@@ -46,10 +50,20 @@ export class EntityTransactionalQueryContext extends EntityQueryContext {
   private readonly postCommitInvalidationCallbacks: PostCommitCallback[] = [];
   private readonly postCommitCallbacks: PostCommitCallback[] = [];
 
+  private readonly preCommitCallbacks: { callback: PreCommitCallback; order: number }[] = [];
+
+  /**
+   * Schedule a pre-commit callback. These will be run within the transaction.
+   * @param callback - callback to schedule
+   * @param order - order in which this should be run, with higher numbers running later than lower numbers
+   */
+  public appendPreCommitCallback(callback: PreCommitCallback, order: number): void {
+    this.preCommitCallbacks.push({ callback, order });
+  }
+
   /**
    * Schedule a post-commit cache invalidation callback. These are run before normal
    * post-commit callbacks in order to have cache consistency in normal post-commit callbacks.
-   *
    * @param callback - callback to schedule
    */
   public appendPostCommitInvalidationCallback(callback: PostCommitCallback): void {
@@ -63,6 +77,17 @@ export class EntityTransactionalQueryContext extends EntityQueryContext {
    */
   public appendPostCommitCallback(callback: PostCommitCallback): void {
     this.postCommitCallbacks.push(callback);
+  }
+
+  public async runPreCommitCallbacksAsync(): Promise<void> {
+    const callbacks = [...this.preCommitCallbacks]
+      .sort((a, b) => a.order - b.order)
+      .map((c) => c.callback);
+    this.preCommitCallbacks.length = 0;
+
+    for (const callback of callbacks) {
+      await callback(this);
+    }
   }
 
   public async runPostCommitCallbacksAsync(): Promise<void> {
--- a/packages/entity/src/EntityQueryContext.ts
+++ b/packages/entity/src/EntityQueryContext.ts
@@ -1,3 +1,5 @@
+import assert from 'assert';
+
 import EntityQueryContextProvider from './EntityQueryContextProvider';
 
 export type PostCommitCallback = (...args: any) => Promise<any>;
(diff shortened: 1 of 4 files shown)


jkhales (approved):
seems useful 🆒

packages/entity/src/EntityQueryContext.ts
```diff
@@ -65,6 +79,17 @@ export class EntityTransactionalQueryContext extends EntityQueryContext {
     this.postCommitCallbacks.push(callback);
   }
 
+  public async runPreCommitCallbacksAsync(): Promise<void> {
```
ide, line 82: IMO it's worth reaffirming one key thing in the documentation: the pre-commit callbacks are run according to their order (should we document that the order is stable otherwise? I'm thinking no, since if you care about the order of callbacks, you should express that with `order`, not with the order in which they are registered).

Relatedly, do you think it is worth grouping callbacks by their order and running each group with Promise.all? (IMO: not that important right now, but may want to leave the API contract open to allow this, in which case we don't want to document that the callbacks are run serially -- just that they are run in order.)

packages/entity/src/__tests__/EntityQueryContext-test.ts
```diff
@@ -0,0 +1,102 @@
...
+      let preCommitSecondCallCount = 0;
+      let postCommitCallCount = 0;
+      let postCommitInvalidationCallCount = 0;
+
+      const preCommitFirstCallback = async (): Promise<void> => {
+        preCommitFirstCallCount++;
+      };
+      const preCommitSecondCallback = async (): Promise<void> => {
+        preCommitSecondCallCount++;
+      };
```
ide, line 23: Could do `const preCommitFirstCallback = jest.fn(async() => {})` and then assert invocation counts with `preCommitFirstCallback.mock.calls.length` or `expect(preCommitFirstCallback).toHaveBeenCalledTimes(n)`

packages/entity/src/EntityQueryContext.ts
```diff
@@ -46,10 +50,20 @@ export class EntityTransactionalQueryContext extends EntityQueryContext {
   private readonly postCommitInvalidationCallbacks: PostCommitCallback[] = [];
   private readonly postCommitCallbacks: PostCommitCallback[] = [];
 
+  private readonly preCommitCallbacks: { callback: PreCommitCallback; order: number }[] = [];
+
+  /**
+   * Schedule a pre-commit callback. These will be run within the transaction.
+   * @param callback - callback to schedule
+   * @param order - order in which this should be run, with higher numbers running later than lower numbers
```
ide, line 58: How do people choose what number to specify for the order? Some guidance would be helpful if the intent is, "I don't care as long as the callback runs before PG's COMMIT" -- should order be `0`? `MIN_SAFE_INTEGER`? `MAX_SAFE_INTEGER`?

packages/entity/src/__tests__/EntityQueryContext-test.ts
```diff
@@ -0,0 +1,102 @@
...
+      const postCommitInvalidationCallback = async (): Promise<void> => {
+        postCommitInvalidationCallCount++;
+        invariant(
+          preCommitFirstCallCount === 1,
+          'preCommitInvalidation should be called before postCommits'
```
ide, line 28:
```suggestion
          'preCommits should be called before postCommitInvalidations'
```

packages/entity/src/__tests__/EntityQueryContext-test.ts
```diff
@@ -0,0 +1,102 @@
...
+        invariant(
+          preCommitSecondCallCount === 1,
+          'preCommitInvalidation should be called before postCommits'
+        );
+      };
+      const postCommitCallback = async (): Promise<void> => {
+        postCommitCallCount++;
+        invariant(
+          preCommitFirstCallCount === 1,
+          'preCommitInvalidation should be called before postCommits'
```
ide, line 39:
```suggestion
          'preCommits should be called before postCommits'
```

packages/entity/src/__tests__/EntityQueryContext-test.ts
```diff
@@ -0,0 +1,102 @@
...
+        );
+        invariant(
+          preCommitSecondCallCount === 1,
+          'preCommitInvalidation should be called before postCommits'
+        );
+        invariant(
+          postCommitInvalidationCallCount === 1,
+          'postCommitInvalidation should be called before postCommit'
```
ide, line 47:
```suggestion
          'postCommitInvalidation should be called before postCommits'
```

(or change "postCommits" to "postCommit" elsewhere)

ide (changes requested):
LGTM but have some minor comments

packages/entity/src/EntityQueryContext.ts
wschurman: Good call. I think it's fine to not group them for now, but I'll update the documentation to indicate that it may be possible in the future.

PR #149, opened 2022-01-30
add new cache flavor to unit test

quinlanj:
# Why

fix failing unit test from https://github.com/expo/universe/pull/9021


wschurman (changes requested):
I think it's probably better if we just re-create `createUnitTestEntityCompanionProvider` in our universe codebase, something like `createExpoUnitTestEntityCompanionProvider`.

PR #152, opened 2022-02-04
[entity][entity-cache-adapter-redis] move transformer responsibility …

quinlanj:
…to cache adapter

# Why

Move the cache object <=> entity object transformation to be the Cache Adapter's responsibility.  Followup to https://github.com/expo/universe/pull/9021#discussion_r794951790

# Test Plan

- [ ] new tests pass

### DIFF
--- a/packages/entity-cache-adapter-redis/src/RedisCacheAdapter.ts
+++ b/packages/entity-cache-adapter-redis/src/RedisCacheAdapter.ts
@@ -4,6 +4,9 @@ import {
   CacheLoadResult,
   FieldTransformerMap,
   mapKeys,
+  transformCacheObjectToFields,
+  CacheStatus,
+  transformFieldsToCacheObject,
 } from '@expo/entity';
 import invariant from 'invariant';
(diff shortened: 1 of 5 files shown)


wschurman (changes requested):
I _think_ we can move the transformation step down one more level to the GenericRedisCacher. This will require making the GenericRedisCacher generic (no pun intended), `GenericRedisCacher<TFields>`. I took a stab at this to make sure it was possible: https://github.com/expo/entity/commit/b7dbc8aabf9e037ee4579d994ced7312416bb3d5 (didn't add tests yet though). It seems like the combination of putting it in the GenericRedisCacher plus changing the `CacheLoadResult` to be generic works.

The only downside I can think of is the generic redis cacher becomes less "generic" in the sense that it can now only cache entities, but I think that's fine. The upside is that we now don't need to reference the redis transformer in the secondary loader (since it makes use of the generic cacher).

Let me know what you think.

quinlanj:
closing in favour of https://github.com/expo/entity/pull/153

PR #153, opened 2022-02-04
move responsibility of transformer to generic redis cacher

quinlanj:
# Why

Move the cache object <=> entity object transformation to be the Generic Redis Cache Adapter's responsibility.  Followup from https://github.com/expo/entity/pull/152#pullrequestreview-873411001

# Test Plan

- [ ] new tests pass

### DIFF
--- a/packages/entity/src/utils/testing/StubCacheAdapter.ts
+++ b/packages/entity/src/utils/testing/StubCacheAdapter.ts
@@ -3,7 +3,6 @@ import invariant from 'invariant';
 import EntityCacheAdapter from '../../EntityCacheAdapter';
 import EntityConfiguration from '../../EntityConfiguration';
 import IEntityCacheAdapterProvider from '../../IEntityCacheAdapterProvider';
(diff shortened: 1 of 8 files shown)


packages/entity/src/utils/testing/StubCacheAdapter.ts
```diff
@@ -53,14 +49,17 @@ export class InMemoryFullCacheStubCacheAdapterProvider implements IEntityCacheAd
...
+      this.cache as Map<string, Readonly<TFields>>
+    );
   }
 }
 
 export class InMemoryFullCacheStubCacheAdapter<TFields> extends EntityCacheAdapter<TFields> {
   constructor(
     entityConfiguration: EntityConfiguration<TFields>,
-    readonly cache: Map<string, Readonly<object>>
+    readonly cache: Map<string, Readonly<TFields>>
```
wschurman, line 62: remove L67-69 below (field transformer stuff). should fix codecov.

wschurman (approved):
Awesome!

PR #154, opened 2022-02-09
v1.0.0

quinlanj:


wschurman (changes requested):
Probably should do v0.23.0 I think. Not quite ready for 1.0.

quinlanj:
I bumped major versions because there are breaking changes if anyone had extended the cache adapter interface in the entity package.

wschurman:
> I bumped major versions because there are breaking changes if anyone had extended the cache adapter interface in the entity package.

Yep, though as I understand semver, pre-1.0.0 the minor version is the breaking change version.

PR #157, opened 2022-02-13
chore: refactor cachers to use an abstract class and interface

quinlanj:
# Why

Interfaces and abstract classes for generic cachers and secondary entity caches. Followup from https://github.com/expo/universe/pull/9106#pullrequestreview-879787342

The plan is to have:

- [GenericRedisCacher ](https://github.com/expo/entity/blob/master/packages/entity-cache-adapter-redis/src/GenericRedisCacher.ts)(from the entity package) and GenericLocalMemoryCacher implement GenericCacher?
- [SecondaryRedisEntityCache](https://github.com/expo/entity/blob/master/packages/entity-secondary-cache-redis/src/RedisSecondaryEntityCache.ts) (from the entity package) and LocalMemoryAndRedisSecondaryEntityCache extending GenericSecondaryEntityCache?

# Test Plan

- [x] current tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+### Chores
+
+* Refactor cachers to use an abstract class and interface  ([#157](https://github.com/expo/entity/pull/157)) 
+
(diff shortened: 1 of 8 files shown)


wschurman (approved):

wschurman:
(codecov looks fine btw)

PR #162, opened 2022-03-08
feat: privacy policy evaluation context

wschurman:
# Why

Just like cascading info was added into triggers in https://github.com/expo/entity/pull/145, this information may be useful when evaluating a privacy policy.

Concretely we have the following use case:
- User has permission to create an object owned by a parent object (user has permission on parent object, privacy policy of child object is derived from permission on parent object). On this object the user's ID is recorded for audit purposes with SET_NULL behavior on the column.
- User creates that object
- User's permission is revoked on parent object but child object is still there (since revoking permission doesn't mean that the data should no longer exist) and therefore still references user.
- User deletes their account, and therefore the SET_NULL deletion behavior is executed. Unfortunately, during this update the user no longer has permission to update the child object and the privacy policy evaluation fails.

# How

This PR adds another piece of information available to privacy policy evaluation, the cascadingDeleteCause. 

This will fix the situation above since we can now check if the update (to SET_NULL) is being executed via a cascade to delete a user, and if so allow in the privacy policy.

# Test Plan

Run all tests.

### DIFF
--- a/packages/entity-example/src/entities/AllowIfUserOwnerPrivacyRule.ts
+++ b/packages/entity-example/src/entities/AllowIfUserOwnerPrivacyRule.ts
@@ -3,6 +3,7 @@ import {
   ReadonlyEntity,
   EntityQueryContext,
   RuleEvaluationResult,
+  EntityPrivacyPolicyEvaluationContext,
 } from '@expo/entity';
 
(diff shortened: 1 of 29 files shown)


quinlanj (approved):
lgtm

ide (approved):

PR #163, opened 2022-03-10
[cache-adapter-local-memory] export classes

quinlanj:
# Why

Actually export classes for use in the package


quinlanj:
closing in favour of https://github.com/expo/entity/pull/164

PR #164, opened 2022-03-10
feat: composed cache adapter, composed secondary cache, local memory secondary cache

wschurman:
# Why

This PR imports some features from Expo's internal testing ground:
- ComposedCacheAdapter - a cache adapter that composes other cache adapters
- ComposedSecondaryCache - a secondary cache that composes other secondary caches
- LocalMemorySecondaryCache - a LRUCache secondary cache (uses local memory stuff from https://github.com/expo/entity/pull/158)

# How

Add features and tests.

# Test Plan

Wait for CI.

### DIFF
--- a/packages/entity-secondary-cache-local-memory/src/LocalMemorySecondaryEntityCache.ts
+++ b/packages/entity-secondary-cache-local-memory/src/LocalMemorySecondaryEntityCache.ts
@@ -1,0 +1,17 @@
+import { GenericSecondaryEntityCache } from '@expo/entity';
+import {
+  GenericLocalMemoryCacher,
+  LocalMemoryCache,
+} from '@expo/entity-cache-adapter-local-memory';
+
+/**
+ * A local memory {@link GenericSecondaryEntityCache}.
+ */
+export default class LocalMemorySecondaryEntityCache<
+  TFields,
+  TLoadParams
+> extends GenericSecondaryEntityCache<TFields, TLoadParams> {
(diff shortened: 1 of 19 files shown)


packages/entity-secondary-cache-local-memory/src/LocalMemorySecondaryEntityCache.ts
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * A local memory {@link GenericSecondaryEntityCache}.
+ */
+export default class LocalMemorySecondaryEntityCache<
+  TFields,
+  TLoadParams
+> extends GenericSecondaryEntityCache<TFields, TLoadParams> {
+  constructor(localMemoryCache: LocalMemoryCache<TFields>) {
+    super(new GenericLocalMemoryCacher(localMemoryCache), (params) => JSON.stringify(params));
```
quinlanj, line 15: I suppose it is possible for `TLoadParams` to contain values that don't stringify well (ie) `[Object object]`. That being said, we do also [make assumptions in the RedisCacheAdapter](https://github.com/expo/entity/blob/master/packages/entity-cache-adapter-redis/src/RedisCacheAdapter.ts#L122) that generics like FieldValue can be coerced to a String. Maybe a good idea to document this in the README or function signature, especially if we document something similar for FieldValue

quinlanj (approved):
thank you!

PR #175, opened 2022-06-08
chore: clean up metrics API a bit and add docblocks

wschurman:
# Why

We don't need three methods for dataloader load count metrics. Instead, just use a type. This way consumers don't need to repeat code.

Also add docblocks in various places.

Closes ENG-5233.

# How

Add type, use `tsc` to ensure it works.

# Test Plan

`yarn tsc`

### DIFF
--- a/packages/entity/src/metrics/IEntityMetricsAdapter.ts
+++ b/packages/entity/src/metrics/IEntityMetricsAdapter.ts
@@ -70,7 +70,7 @@ export enum IncrementLoadCountEventType {
   CACHE,
 
   /**
-   * Type for when a database load is initiated due to a dataloader and cache miss.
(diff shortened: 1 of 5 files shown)


packages/entity/src/metrics/IEntityMetricsAdapter.ts
```diff
@@ -23,13 +41,57 @@ export enum EntityMetricsMutationType {
...
+   */
+  DATALOADER,
+
+  /**
+   * Type for when a cache load is initiated due to a dataloader miss.
+   */
+  CACHE,
+
+  /**
+   * Type for when a database load is initiated due to a dataloader and cache miss.
```
ide, line 73:
```suggestion
   * Type for when a database load is initiated due to a dataloader and cache miss, when an entity query doesn't support caching, or during a transaction.
```

ide (approved):

PR #177, opened 2022-07-01
feat: add EntityField transformation

wschurman:
# Why

This is a first attempt at field transformation (right now we just do field validation). Note that this is currently broken due to the following case:

1. Entity contains LowerCaseStringField
2. `loadManyByFieldEqualingMany('lowercaseField', 'THIS_WILL_BE_LOWERCASED')` returns a map from value requested to entities matching that value. The problem is that the value requested is upper-case but the returned map is lower-cased.

# How

Do transformation in addition to validation in the field.

# Test Plan

Run all tests.


wschurman:
Abandoning, need for this has dwindled.

PR #179, opened 2022-07-14
fix: fix workflow branch specification

wschurman:
# Why

The main branch was changed from `master` to `main`, but the CI wasn't updated. This is causing codecov and CI not to run on the main branch.

# How

Update the workflow.

# Test Plan

Wait for CI.


ide (commented):
✅

PR #182, opened 2022-07-27
[fix/feat] recache values in higher layers for composed cache adapter

quinlanj:
# Why
This PR recaches the values of lower cache layers missing from the higher layers of a ComposedCacheAdapter.

Consider a ComposedCacheAdapter with a local memory cache and a redis cache. In www, values typically stay in Redis much longer than they stay in Local Memory (capped to 5 seconds). In the scenario that a value is already cached in Redis but no longer cached in Local Memory (either it expired or was not the machine that originally cached to Redis), we will consider this a cache hit. Because the only time we cache values is when we fetch from DB, we will continually fetch from Redis until the value is evicted, at which point we will cache back into all the layers.

A better system would be to recache the higher layers with values found in the lower layers so we aren't loading from Redis most of the time. 

# How

recache higher layers in `loadManyAsync`

# Test Plan

- [ ] added new test

### DIFF
--- a/packages/entity/src/ComposedEntityCacheAdapter.ts
+++ b/packages/entity/src/ComposedEntityCacheAdapter.ts
@@ -60,7 +60,7 @@ export default class ComposedEntityCacheAdapter<TFields> extends EntityCacheAdap
     for (let i = this.cacheAdapters.length - 1; i >= 0; i--) {
       const cacheAdapter = nullthrows(this.cacheAdapters[i]);
       const hitsToCache = new Map<NonNullable<TFields[N]>, Readonly<TFields>>();
-      const missesToCache: NonNullable<TFields[N]>[] = [];
+      const negativesToCache: NonNullable<TFields[N]>[] = [];
 
       // Loop over all lower layer caches to collect hits and misses
       for (let j = i + 1; j < this.cacheAdapters.length; j++) {
@@ -70,7 +70,7 @@ export default class ComposedEntityCacheAdapter<TFields> extends EntityCacheAdap
           if (cacheResult.status === CacheStatus.HIT) {
             hitsToCache.set(fieldValue, cacheResult.item);
           } else if (cacheResult.status === CacheStatus.NEGATIVE) {
-            missesToCache.push(fieldValue);
+            negativesToCache.push(fieldValue);
           }
         });
       }
@@ -79,8 +79,8 @@ export default class ComposedEntityCacheAdapter<TFields> extends EntityCacheAdap
       if (hitsToCache.size > 0) {
         promises.push(cacheAdapter.cacheManyAsync(fieldName, hitsToCache));
       }
-      if (missesToCache.length > 0) {
-        promises.push(cacheAdapter.cacheDBMissesAsync(fieldName, missesToCache));
+      if (negativesToCache.length > 0) {
+        promises.push(cacheAdapter.cacheDBMissesAsync(fieldName, negativesToCache));
       }
       await Promise.all(promises);
     }
--- a/packages/entity/src/ComposedEntityCacheAdapter.ts
+++ b/packages/entity/src/ComposedEntityCacheAdapter.ts
@@ -27,7 +27,8 @@ export default class ComposedEntityCacheAdapter<TFields> extends EntityCacheAdap
     fieldValues: readonly NonNullable<TFields[N]>[]
   ): Promise<ReadonlyMap<NonNullable<TFields[N]>, CacheLoadResult<TFields>>> {
     const retMap = new Map<NonNullable<TFields[N]>, CacheLoadResult<TFields>>();
-    const fulfilledFieldValuesByCacheIndex: NonNullable<TFields[N]>[][] = this.cacheAdapters.map(
+    const fulfilledFieldValuesByCacheIndex: NonNullable<TFields[N]>[][] = Array.from(
+      { length: this.cacheAdapters.length },
       () => []
     );
 
--- a/packages/entity/src/ComposedEntityCacheAdapter.ts
+++ b/packages/entity/src/ComposedEntityCacheAdapter.ts
@@ -44,8 +44,7 @@ export default class ComposedEntityCacheAdapter<TFields> extends EntityCacheAdap
           newUnfulfilledFieldValues.push(fieldValue);
         } else {
           retMap.set(fieldValue, cacheResult);
-          const fulfilledFieldValues = nullthrows(fulfilledFieldValuesByCacheIndex[i]);
-          fulfilledFieldValues.push(fieldValue);
+          nullthrows(fulfilledFieldValuesByCacheIndex[i]).push(fieldValue);
         }
       }
       unfulfilledFieldValues = newUnfulfilledFieldValues;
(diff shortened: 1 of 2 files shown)


packages/entity/src/ComposedEntityCacheAdapter.ts
```diff
@@ -40,11 +44,35 @@ export default class ComposedEntityCacheAdapter<TFields> extends EntityCacheAdap
           newUnfulfilledFieldValues.push(fieldValue);
         } else {
           retMap.set(fieldValue, cacheResult);
+          const fulfilledFieldValues = nullthrows(fulfilledFieldValuesByCacheIndex[i]);
+          fulfilledFieldValues.push(fieldValue);
```
wschurman, line 48:
```suggestion
          nullthrows(fulfilledFieldValuesByCacheIndex[i]).push(fieldValue);
```
(makes it clearer that this is a object reference rather than a local variable)

packages/entity/src/ComposedEntityCacheAdapter.ts
```diff
@@ -40,11 +44,35 @@ export default class ComposedEntityCacheAdapter<TFields> extends EntityCacheAdap
...
         }
       }
       unfulfilledFieldValues = newUnfulfilledFieldValues;
     }
 
+    // Recache values from higher layers that were not found in lower layers
+    for (let i = 0; i < this.cacheAdapters.length; i++) {
```
wschurman, line 55: we should be able to run a `Promise.all` here instead of awaiting within a for loop since the re-cache operations are logically independent

packages/entity/src/ComposedEntityCacheAdapter.ts
```diff
@@ -26,9 +26,13 @@ export default class ComposedEntityCacheAdapter<TFields> extends EntityCacheAdap
     fieldValues: readonly NonNullable<TFields[N]>[]
   ): Promise<ReadonlyMap<NonNullable<TFields[N]>, CacheLoadResult<TFields>>> {
     const retMap = new Map<NonNullable<TFields[N]>, CacheLoadResult<TFields>>();
+    const fulfilledFieldValuesByCacheIndex: NonNullable<TFields[N]>[][] = this.cacheAdapters.map(
+      () => []
+    );
 
     let unfulfilledFieldValues = fieldValues;
-    for (const cacheAdapter of this.cacheAdapters) {
+    for (let i = 0; i < this.cacheAdapters.length; i++) {
```
wschurman, line 34: We await within this loop to fetch less objects each time, though might it actually be faster to just fetch all from all caches in parallel as well and then merge them? Might be worth profiling. If we do stick with the for loop  we should probably have an early break when `unfulfilledFieldValues` is empty.

packages/entity/src/ComposedEntityCacheAdapter.ts
```diff
@@ -40,11 +44,35 @@ export default class ComposedEntityCacheAdapter<TFields> extends EntityCacheAdap
...
       }
       unfulfilledFieldValues = newUnfulfilledFieldValues;
     }
 
+    // Recache values from higher layers that were not found in lower layers
+    for (let i = 0; i < this.cacheAdapters.length; i++) {
+      const cacheAdapter = nullthrows(this.cacheAdapters[i]);
+      const hitsToCache = new Map<NonNullable<TFields[N]>, Readonly<TFields>>();
+      const missesToCache: NonNullable<TFields[N]>[] = [];
+      for (let j = i + 1; j < this.cacheAdapters.length; j++) {
```
wschurman, line 59: nit: add more inline comments about what this and the outer for loop index math is doing

packages/entity/src/ComposedEntityCacheAdapter.ts
quinlanj: Performance here will depend on the type of caches that are in the ComposedEntityCacheAdapter. In our case where we've got a local memory and redis cache, I would strongly prefer to keep the for-loop.  

If we fetched everything in parallel, we will be waiting for the slowest cache every single time (redis). In the case we have a popular query and the caches are warm, we'll be waiting on the redis cache even though all the values are available in local memory. This would be unfortunate because the redis cache is magnitudes slower than the local memory cache.

The for-loop will optimize for the case when the cache is warm (which will be the majority of requests for a popular query), since we will only be waiting on the local memory cache

wschurman: Good call. Yeah that makes sense.

We should still be able to get away with an early break out of the loop if `unfulfilledFieldValues` is empty though I think.

packages/entity/src/ComposedEntityCacheAdapter.ts
quinlanj: looking at the other code, we've made a point to `await` on caches closest to the database, then move onto the ones closer to the application. I think this was to avoid bugs where the cache closest to the application was stale, but the ones closer to the application was more up-to-date. 

I think the most correct approach would be to iterate backwards (ie) `for (let i = this.cacheAdapters.length - 1; i >= 0; i--) {` to keep this behaviour consistent

wschurman: Ah right. Good call.

packages/entity/src/ComposedEntityCacheAdapter.ts
```diff
@@ -26,9 +27,13 @@ export default class ComposedEntityCacheAdapter<TFields> extends EntityCacheAdap
     fieldValues: readonly NonNullable<TFields[N]>[]
   ): Promise<ReadonlyMap<NonNullable<TFields[N]>, CacheLoadResult<TFields>>> {
     const retMap = new Map<NonNullable<TFields[N]>, CacheLoadResult<TFields>>();
+    const fulfilledFieldValuesByCacheIndex: NonNullable<TFields[N]>[][] = this.cacheAdapters.map(
+      () => []
+    );
```
wschurman, line 32: optional: I've started seeing this syntax around for creating an array of specific length with initialized values. not sure what the relative performance is.

```suggestion
    const fulfilledFieldValuesByCacheIndex: NonNullable<TFields[N]>[][] = Array.from(
      { length: this.cacheAdapters.length },
      () => []
    );
```

packages/entity/src/ComposedEntityCacheAdapter.ts
```diff
@@ -40,9 +45,43 @@ export default class ComposedEntityCacheAdapter<TFields> extends EntityCacheAdap
...
+        break;
+      }
+    }
+
+    // Recache values from lower layers that were not found in higher layers
+    // Write to lower layers first
+    for (let i = this.cacheAdapters.length - 1; i >= 0; i--) {
+      const cacheAdapter = nullthrows(this.cacheAdapters[i]);
+      const hitsToCache = new Map<NonNullable<TFields[N]>, Readonly<TFields>>();
+      const missesToCache: NonNullable<TFields[N]>[] = [];
```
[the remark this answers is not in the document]
wschurman: (or `dbMissesToCache` up to you)

packages/entity/src/__tests__/ComposedCacheAdapter-test.ts
```diff
@@ -183,6 +183,31 @@ describe(ComposedEntityCacheAdapter, () => {
...
+      ]);
+
+      expect(primaryResults.get('test-id-1')).toMatchObject({
+        status: CacheStatus.HIT,
+        item: { id: 'test-id-1' },
+      });
+      expect(primaryResults.get('test-id-2')).toMatchObject({ status: CacheStatus.NEGATIVE });
+      expect(primaryResults.get('test-id-3')).toMatchObject({ status: CacheStatus.MISS });
+      expect(primaryResults.size).toBe(3);
+    });
```
wschurman, line 209: could be a good sanity check to do a subsequent load from `cacheAdapter` after the first load to ensure that these writes don't change the output

wschurman (approved):
Just a few minor changes, lgtm otherwise though.

PR #183, opened 2022-07-28
v0.26.1

quinlanj:
# Why
Releasing.

# How
Following instructions in readme.

# Test Plan
Wait for CI


wschurman (approved):
Looks like the change didn't get picked up automatically in the changelogs since it didn't follow the format of conventional commits. That's okay though. It doesn't matter too much. Could be good to add a manual changelog entry though in this PR.

PR #184, opened 2022-07-28
docs: update readme deployment

quinlanj:
readme nits


wschurman (approved):
double check that the format of the commit title adheres to conventional commit style when you land this. maybe we should add a commit hook or something.

PR #185, opened 2022-08-01
feat: add orderByRaw to loadManyByRawWhereClauseAsync

FiberJW:
# Why

To achieve the sort I want for Expo `apps` (`ORDER BY GREATEST(display_name, SPLIT_PART(full_name, '/', 2)) ASC`), Entity needs a raw order clause option.

# How

I added an `orderByRaw` property to the `querySelectionModifiers` for `loadManyByRawWhereClauseAsync` and updated all of the downstream types and call sites accordingly.

# Test Plan

The added integration test and Typescript checks should pass.

### DIFF
--- a/packages/entity-database-adapter-knex/src/PostgresEntityDatabaseAdapter.ts
+++ b/packages/entity-database-adapter-knex/src/PostgresEntityDatabaseAdapter.ts
@@ -7,6 +7,7 @@ import {
   TableQuerySelectionModifiers,
   TableFieldSingleValueEqualityCondition,
   TableFieldMultiValueEqualityCondition,
+  TableQuerySelectionModifiersWithOrderByRaw,
 } from '@expo/entity';
 import { Knex } from 'knex';
 
@@ -53,11 +54,25 @@ export default class PostgresEntityDatabaseAdapter<TFields> extends EntityDataba
     );
   }
 
+  private applyQueryModifiersToQueryOrderByRaw(
+    query: Knex.QueryBuilder,
+    querySelectionModifiers: TableQuerySelectionModifiersWithOrderByRaw
+  ): Knex.QueryBuilder {
+    let ret = this.applyQueryModifiersToQuery(query, querySelectionModifiers);
+
+    const { orderByRaw } = querySelectionModifiers;
+    if (orderByRaw !== undefined) {
+      ret = ret.orderByRaw(orderByRaw);
+    }
+
+    return ret;
+  }
+
   private applyQueryModifiersToQuery(
     query: Knex.QueryBuilder,
     querySelectionModifiers: TableQuerySelectionModifiers
   ): Knex.QueryBuilder {
-    const { orderBy, offset, limit, orderByRaw } = querySelectionModifiers;
+    const { orderBy, offset, limit } = querySelectionModifiers;
 
     let ret = query;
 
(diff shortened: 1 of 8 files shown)


packages/entity-database-adapter-knex/src/__integration-tests__/PostgresEntityIntegration-test.ts
```diff
@@ -233,6 +233,53 @@ describe('postgres entity integration', () => {
...
+        PostgresTestEntity.loader(vc1)
+          .enforcing()
+          .loadManyByFieldEqualityConjunctionAsync([], {
+            orderByRaw: [
+              {
+                clause: 'invalid_column DESC',
+              },
+            ],
+          })
+      ).rejects.toThrow();
```
wschurman, line 280: We should probably be explicit about what error we expect this to throw. In this case I believe it throws a `EntityDatabaseAdapterUnknownError`, which I think is okay for now (probably doesn't make sense to create a new error type for this type of error right now).

packages/entity-database-adapter-knex/src/PostgresEntityDatabaseAdapter.ts
```diff
@@ -67,6 +67,12 @@ export default class PostgresEntityDatabaseAdapter<TFields> extends EntityDataba
       }
     }
 
+    if (orderByRaw !== undefined) {
+      for (const orderByRawSpecification of orderByRaw) {
```
wschurman, line 71: I'm thinking that we probably don't want this to be an array, though I could be convinced otherwise. The reason is that even though knex supports it, we don't support multiple `whereRaw` in `fetchManyByRawWhereClauseInternalAsync`, so we should keep them in alignment.

wschurman (changes requested):
Looks good! Just a few things to add:
- In `StubDatabaseAdapter`, we should throw an exception in `fetchManyByFieldEqualityConjunctionInternalAsync` if `orderByRaw` is specified. Same change should be applied to `InMemoryDatabaseAdapter`.

One larger question: Do you think this makes more sense to limit this arg to `loadManyByRawWhereClauseAsync`? Having it present in `loadManyByFieldEqualityConjunctionAsync` creates a situation where we can't simulate a query in memory (see note above about adding the throws) where up until now we could. My gut says that so far we've used `loadManyByRawWhereClauseAsync` as the escape hatch for needing to do special SQL things in the database adapters, so it might make more sense for this to just be available in that method (especially because we can't simulate it in the in-memory database adapters mentioned above). That being said, I think your particular use case used `loadManyByFieldEqualityConjunctionAsync` so maybe not?

FiberJW:
I think it makes sense to limit this arg to `loadManyByRawWhereClauseAsync`.

Which of the following do you think is the better approach:
- error out in `loadManyByFieldEqualityConjunctionAsync` when `orderByRaw` is provided in `querySelectionModifiers`
- add a new `orderByRaw` param to `loadManyByRawWhereClauseAsync` after `querySelectionModifiers`
- extend the `querySelectionModifiers` type like this PR does, but only in the `loadManyByRawWhereClauseAsync` codepath (`QuerySelectionModifiers<TFields> & { orderByRaw?: string }`)

wschurman:
> * extend the `querySelectionModifiers` type like this PR does, but only in the `loadManyByRawWhereClauseAsync` codepath (`QuerySelectionModifiers<TFields> & { orderByRaw?: string }`)

I like this option.

wschurman (changes requested):
Looks a lot better. Probably worth create a new set of types that extends the query modifiers type that includes the new param. This way the internal and external types of the library match and are named. I tried it in https://github.com/expo/entity/commit/3d2d1898afb5db42b3c1197541fe0127f5813219 to be sure and it seemed to work (feel free to use all or none of that commit).

wschurman (approved):

wschurman:
(don't worry about diff target codecov percentage)

wschurman:
Ping me when you land this and I'll create a new release of these and upgrade them in www.

PR #193, opened 2022-08-31
feat: add BigIntField

wschurman:
# Why

Noticed by @FiberJW in https://github.com/expo/universe/pull/10420#discussion_r953010810. Knex (node-postgres) represents the BIGINT type as a string, so using `IntField` is incorrect. 

# How

This fixes that by adding a `BigInt` field that is of string type.

# Test Plan

Run new test case.

### DIFF
--- a/packages/entity-database-adapter-knex/src/__tests__/EntityFields-test.ts
+++ b/packages/entity-database-adapter-knex/src/__tests__/EntityFields-test.ts
@@ -1,0 +1,14 @@
+import { describeFieldTestCase } from '@expo/entity';
+
+import { JSONArrayField, MaybeJSONArrayField } from '../EntityFields';
+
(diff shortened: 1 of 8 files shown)


FiberJW (approved):
This looks good to me! Thanks for updating the library to support my GitHub integration work!

thumbs-up-00037

packages/entity-database-adapter-knex/src/__tests__/EntityFields-test.ts
```diff
@@ -0,0 +1,14 @@
+import { describeFieldTestCase } from '@expo/entity';
+
+import { JSONArrayField, MaybeJSONArrayField } from '../EntityFields';
+
```
wschurman, line 4: this needs a test case for BigIntField

PR #196, opened 2022-10-04
feat: add sharded redis cache adapter

wschurman:
# Why

This PR implements a sharded Redis cache adapter, where cache keys are spread evenly across a user-configurable set of redis instances.

Due to the user-configurable nature of this, any [consistent hashing](https://dgryski.medium.com/consistent-hashing-algorithmic-tradeoffs-ef6b8e2fcae8) scheme and configuration can be employed by the library consumer. One good one might be `xxhashjs` when used in production.

Due to the short-lived nature of most entity cache entries (and the fact that they are cache and not persisted data), changing the number of shards is a bit easier since we can just change the cache key at the same time. This PR provides a mechanism for doing that.

# How

1. Clean up the `GenericEntityCacheAdapter` concept (move it into the core package).
2. Introduce new ShardedGenericRedisCacher.

In a follow-up PR, we can make the GenericRedisCacher the n=1 case of ShardedGenericRedisCacher to reduce code duplication.

# Test Plan

Run all tests.


wschurman:
One alternative simpler version of this might be the following:

- Change the entity redis cache adapter to operate on a Redis-like interface
- Create a new object, ShardedRedis that implements that interface and decides which instance to send it to

The upside here is that it is less code most likely. The downside is that it muddies the redis `multi` transaction a little bit (no longer transactional for all keys, though this PR also has that downside).

wschurman:
Moving this back to draft as we'll be going with the other option for now.

PR #197, opened 2022-10-05
feat: alternative sharded redis cache

wschurman:
# Why

This is the alternative to https://github.com/expo/entity/pull/196 mentioned in https://github.com/expo/entity/pull/196#issuecomment-1267750484.

# How

Instead of doing the sharding logic in the entity cache adapter level, do it within the redis interface. This requires some bookkeeping and tricks within the multi-redis logic, but isn't horrendous.

The main thing to review is the `ShardedRedis` class in the test. This is what would need to be duplicated to an application that uses it (though with a real consistent hash).

# Test Plan

Run test (same test as other PR to prove it works).

### DIFF
--- a/packages/entity-cache-adapter-redis/src/GenericRedisCacher.ts
+++ b/packages/entity-cache-adapter-redis/src/GenericRedisCacher.ts
@@ -6,7 +6,6 @@ import {
   transformFieldsToCacheObject,
   IEntityGenericCacher,
 } from '@expo/entity';
-import { Redis } from 'ioredis';
 
 import { redisTransformerMap } from './RedisCommon';
 import wrapNativeRedisCallAsync from './errors/wrapNativeRedisCallAsync';
@@ -15,11 +14,22 @@ import wrapNativeRedisCallAsync from './errors/wrapNativeRedisCallAsync';
 // The sentinel value is distinct from any (positively) cached value.
 const DOES_NOT_EXIST_REDIS = '';
 
+export interface IRedisTransaction {
+  set(key: string, value: string, secondsToken: 'EX', seconds: number): this;
+  exec(): Promise<any>;
+}
+
+export interface IRedis {
+  mget(...args: [...keys: string[]]): Promise<(string | null)[]>;
+  multi(): IRedisTransaction;
+  del(...args: [...keys: string[]]): Promise<any>;
+}
+
 export interface GenericRedisCacheContext {
   /**
    * Instance of ioredis.Redis
    */
-  redisClient: Redis;
+  redisClient: IRedis;
 
   /**
    * TTL for caching database hits. Successive entity loads within this TTL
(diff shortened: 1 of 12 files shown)


packages/entity-cache-adapter-redis/src/GenericRedisCacher.ts
```diff
@@ -15,11 +14,22 @@ import wrapNativeRedisCallAsync from './errors/wrapNativeRedisCallAsync';
...
 
+export interface IRedisTransaction {
+  set(key: string, value: string, secondsToken: 'EX', seconds: number): this;
+  exec(): Promise<any>;
+}
+
+export interface IRedis {
+  mget(...args: [...keys: string[]]): Promise<(string | null)[]>;
+  multi(): IRedisTransaction;
+  del(...args: [...keys: string[]]): Promise<any>;
```
ide, line 25:
```suggestion
  del(...args: [...keys: string[]]): Promise<number>;
```
> [Integer reply](https://redis.io/docs/reference/protocol-spec#resp-integers): The number of keys that were removed.

packages/entity-cache-adapter-redis/src/__integration-tests__/ShardedRedisCacheAdapter-integration-test.ts
```diff
@@ -0,0 +1,218 @@
...
+    const redisClientsForKeys = this.getRedisClientsForKeys(args);
+    await mapMapAsync(redisClientsForKeys, async ({ keys, redisClient }) => {
+      await redisClient.del(...keys);
+    });
+  }
+
+  private getRedisClientsForKeys(
+    keys: readonly string[]
+  ): Map<number, { keys: string[]; redisClient: Redis }> {
+    const shardGroupsForKeys = new Map(keys.map((k) => [k, getShardForKey(k)]));
```
ide, line 101: Instead of shard groups, aren't these just shards? (Map<Key, Shard>)

packages/entity-cache-adapter-redis/src/__integration-tests__/ShardedRedisCacheAdapter-integration-test.ts
```diff
@@ -0,0 +1,218 @@
...
+      const entry = computeIfAbsent(redisClientsForKeys, shardGroup, (currShardGroup) => ({
+        keys: [],
+        redisClient: this.getRedisInstanceForShardGroup(currShardGroup),
+      }));
+      entry.keys.push(key);
+    }
+    return redisClientsForKeys;
+  }
+
+  private getRedisInstanceForShardGroup(shardGroup: number): Redis {
```
ide, line 113: The other places in the code refer to "Redis clients":
```suggestion
  private getRedisClientForShardGroup(shardGroup: number): Redis {
```

packages/entity-cache-adapter-redis/src/__integration-tests__/ShardedRedisCacheAdapter-integration-test.ts
```diff
@@ -0,0 +1,218 @@
...
+        return ret;
+      },
+      async exec(): Promise<any> {
+        return await parentExec();
+      },
+    };
+    return ret;
+  }
+
+  async del(...args: string[]): Promise<void> {
```
ide, line 91: In a non-test implementation, this should return a number (the sum of the individual del results)

ide (approved):

wschurman:
This was a proof of concept. No longer needed for now.

PR #199, opened 2022-10-05
feat: add example of redis mget batching using @expo/batcher

wschurman:
# Why

Add example of how to coalesce mgets within the redis interface. This will allow multiple entity companion instances (multiple requests) to more efficiently fetch values from redis.

# How

Use `@expo/batcher` to batch `mgets`. The other `IRedis` methods are not batched and probably don't need to be since they aren't used as frequently.

# Test Plan

Run new test.

### DIFF
--- a/packages/entity-cache-adapter-redis/src/__integration-tests__/BatchedRedisCacheAdapter-integration-test.ts
+++ b/packages/entity-cache-adapter-redis/src/__integration-tests__/BatchedRedisCacheAdapter-integration-test.ts
@@ -1,0 +1,164 @@
+import { Batcher } from '@expo/batcher';
+import { ViewerContext, zipToMap } from '@expo/entity';
+import invariant from 'invariant';
+import Redis from 'ioredis';
+import nullthrows from 'nullthrows';
+import { URL } from 'url';
+import { v4 as uuidv4 } from 'uuid';
+
+import { IRedis, IRedisTransaction } from '../GenericRedisCacher';
+import RedisCacheAdapter, { RedisCacheAdapterContext } from '../RedisCacheAdapter';
+import RedisTestEntity from '../testfixtures/RedisTestEntity';
+import { createRedisIntegrationTestEntityCompanionProvider } from '../testfixtures/createRedisIntegrationTestEntityCompanionProvider';
+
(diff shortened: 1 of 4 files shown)


packages/entity-cache-adapter-redis/src/__integration-tests__/BatchedRedisCacheAdapter-integration-test.ts
```diff
@@ -0,0 +1,164 @@
...
+import { IRedis, IRedisTransaction } from '../GenericRedisCacher';
+import RedisCacheAdapter, { RedisCacheAdapterContext } from '../RedisCacheAdapter';
+import RedisTestEntity from '../testfixtures/RedisTestEntity';
+import { createRedisIntegrationTestEntityCompanionProvider } from '../testfixtures/createRedisIntegrationTestEntityCompanionProvider';
+
+class BatchedRedis implements IRedis {
+  private readonly mgetBatcher = new Batcher<string[], (string | null)[]>(
+    this.batchMgetAsync.bind(this),
+    {
+      maxBatchInterval: 0,
```
ide, line 18: In the future, we might want to set a max batch size so that requests and responses don't become too big, though ideally we'd compute this size based on the number of unique keys to fetch which would require changes to the batcher API (e.g. a function that is passed in that the batcher calls to determine whether to run)

ide (approved):
Making the batcher tunable might be useful but shouldn't block this PR. We should wait until we see if batching helps at all.

Might want to add a test case where two entities are deleted at the same time. There should be one batched DEL sent to Redis, though the timing is tricky because both Postgres deletions need to complete during the same microtask tick, so maybe that's better as a unit test with more mocking.

wschurman:
> Making the batcher tunable might be useful but shouldn't block this PR. We should wait until we see if batching helps at all.

Yep. This PR is mostly just an example for this open source repo. We'll do the actual implementation in the expo backend repo.

> Might want to add a test case where two entities are deleted at the same time. There should be one batched DEL sent to Redis, though the timing is tricky because both Postgres deletions need to complete during the same microtask tick, so maybe that's better as a unit test with more mocking.

We don't batch the `del` in this example. We can discuss batching it in the expo backend implementation PR.

PR #203, opened 2022-12-23
[www] Upgrade knex to 2.3.0

wschurman:
# Why

Trying to upgrade this where we use it, but it is resulting in conflicting typescript types I believe. https://github.com/expo/universe/pull/11100

https://nvd.nist.gov/vuln/detail/CVE-2016-20018 says Knex has a SQL injection that can cause the WHERE clause to be ignored. It's unclear how this is triggered, and due to the usage pattern within entity it is doubtful that it affects us, but still good to stay upgraded.

How
---
Upgrade to Knex 2.3.0. This is a major semver upgrade. However, the breaking change in 2.3.0 was adding back support for sqlite3. That does not affect us. I read through the changelog and saw no other changes that would affect us, though there is better support for JSONB and Postgres now: https://github.com/knex/knex/pull/5201

# Test Plan

Run all tests.

### DIFF
--- a/packages/entity-database-adapter-knex/package.json
+++ b/packages/entity-database-adapter-knex/package.json
@@ -30,7 +30,7 @@
     "@expo/entity": "*"
   },
   "dependencies": {
-    "knex": "^1.0.2"
+    "knex": "^2.3.0"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


ide (approved):
Thanks!

PR #206, opened 2023-01-24
[rfc] loadFirstByFieldEqualityConjunction, a convenience method

ide:
# Why

Several places (about five) in our codebase call `loadFirstByFieldEqualityConjunction` with `{ limit: 1 }` either to get just the first row or to test for the existence of any matching row. The intent of the callers is often to get the latest row (limit 1, order by a descending timestamp). This PR adds a convenience method for that use case, primarily to help the callers convey their intent better, rather than saving one line of code by removing the need to specify `limit: 1`.

# How

Added `loadFirstByFieldEqualityConjunction` to EntityLoader and EnforcingEntityLoader. This new method calls into `loadManyByFieldEqualityConjunction` and passes in `limit: 1`. It uses TS to prevent `limit` from being specified from the outside, which indicates a programming error. It also requires `orderBy` to be specified. We have a caller that doesn't care about order, just existence -- in this case, the caller should pass in `orderBy: []` to explicitly show that it doesn't care about order. The more common pattern was to get the first row of an ordered set, though.

# Test Plan

Added unit tests for both EntityLoader and EnforcingEntityLoader. Notably, the EntityLoader test ensures the privacy policy was evaluated only once even though two rows match, and also checks that the `orderBy` clause was honored.

### DIFF
--- a/packages/entity/src/EnforcingEntityLoader.ts
+++ b/packages/entity/src/EnforcingEntityLoader.ts
@@ -118,6 +118,22 @@ export default class EnforcingEntityLoader<
    * Enforcing version of entity loader method by the same name.
    * @throws EntityNotAuthorizedError when viewer is not authorized to view one or more of the returned entities
    */
+  async loadFirstByFieldEqualityConjunctionAsync<N extends keyof Pick<TFields, TSelectedFields>>(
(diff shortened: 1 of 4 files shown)


wschurman (approved):
I think this sounds reasonable.

PR #207, opened 2023-04-19
feat: add ability to specify knex transaction config

wschurman:
# Why

Expo has a need for serializable transaction isolation level. This change allows specifying this configuration for knex.

# How

Add transaction config, pass through (with translation particular to the adaptor flavor).

# Test Plan

Run new tests.

### DIFF
--- a/packages/entity/src/EntityQueryContext.ts
+++ b/packages/entity/src/EntityQueryContext.ts
@@ -9,11 +9,9 @@ export type PreCommitCallback = (
 ) => Promise<any>;
 
 export enum TransactionIsolationLevel {
-  READ_UNCOMMITTED,
-  READ_COMMITTED,
-  SNAPSHOT,
-  REPEATABLE_READ,
-  SERIALIZABLE,
+  READ_COMMITTED = 'READ_COMMITTED',
+  REPEATABLE_READ = 'REPEATABLE_READ',
+  SERIALIZABLE = 'SERIALIZABLE',
 }
 
 export type TransactionConfig = {
@@ -154,7 +152,7 @@ export class EntityTransactionalQueryContext extends EntityQueryContext {
   ): Promise<T> {
     assert(
       transactionConfig === undefined,
-      'should not pass transactionConfig to an already created transaction'
+      'Should not pass transactionConfig to a nested transaction'
     );
     return await transactionScope(this);
   }
--- a/packages/entity/src/EntityQueryContext.ts
+++ b/packages/entity/src/EntityQueryContext.ts
@@ -8,6 +8,18 @@ export type PreCommitCallback = (
   ...args: any
 ) => Promise<any>;
 
+export enum TransactionIsolationLevel {
+  READ_UNCOMMITTED,
+  READ_COMMITTED,
+  SNAPSHOT,
+  REPEATABLE_READ,
+  SERIALIZABLE,
+}
+
+export type TransactionConfig = {
+  isolationLevel?: TransactionIsolationLevel;
+};
+
 /**
  * Entity framework representation of transactional and non-transactional database
  * query execution units.
@@ -25,7 +37,8 @@ export abstract class EntityQueryContext {
   }
 
   abstract runInTransactionIfNotInTransactionAsync<T>(
-    transactionScope: (queryContext: EntityTransactionalQueryContext) => Promise<T>
+    transactionScope: (queryContext: EntityTransactionalQueryContext) => Promise<T>,
+    transactionConfig?: TransactionConfig
   ): Promise<T>;
 }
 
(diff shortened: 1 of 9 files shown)


packages/entity/src/EntityQueryContextProvider.ts
```diff
@@ -23,24 +24,26 @@ export default abstract class EntityQueryContextProvider {
...
   ): (transactionScope: (queryInterface: any) => Promise<T>) => Promise<T>;
 
   /**
    * Start a transaction and execute the provided transaction-scoped closure within the transaction.
    * @param transactionScope - async callback to execute within the transaction
    */
   async runInTransactionAsync<T>(
-    transactionScope: (queryContext: EntityTransactionalQueryContext) => Promise<T>
+    transactionScope: (queryContext: EntityTransactionalQueryContext) => Promise<T>,
+    transactionConfig?: TransactionConfig
```
ide, line 42: suggestion to make this default to an empty object so that the rest of the methods can take a non-nullable type:
```suggestion
    transactionConfig: TransactionConfig = {}
```

packages/entity/src/EntityQueryContextProvider.ts
```diff
@@ -58,11 +61,15 @@ export default abstract class EntityQueryContextProvider {
    */
   async runInNestedTransactionAsync<T>(
     outerQueryContext: EntityTransactionalQueryContext,
-    transactionScope: (innerQueryContext: EntityNestedTransactionalQueryContext) => Promise<T>
+    transactionScope: (innerQueryContext: EntityNestedTransactionalQueryContext) => Promise<T>,
+    transactionConfig?: TransactionConfig
```
ide, line 65: Should this just not have a transactionConfig parameter?

packages/entity/src/EntityQueryContextProvider.ts
```diff
@@ -23,24 +24,26 @@ export default abstract class EntityQueryContextProvider {
...
-    transactionScope: (queryInterface: any) => Promise<T>
-  ) => Promise<T>;
+  protected abstract createTransactionRunner<T>(
+    transactionConfig?: TransactionConfig
+  ): (transactionScope: (queryInterface: any) => Promise<T>) => Promise<T>;
 
   protected abstract createNestedTransactionRunner<T>(
-    outerQueryInterface: any
+    outerQueryInterface: any,
+    transactionConfig?: TransactionConfig
```
ide, line 33: Is this parameter needed? Do we want to support this since Postgres doesn't support nested transactions (just savepoints)?

packages/entity/src/EntityQueryContext.ts
```diff
@@ -132,8 +149,13 @@ export class EntityTransactionalQueryContext extends EntityQueryContext {
   }
 
   async runInTransactionIfNotInTransactionAsync<T>(
-    transactionScope: (queryContext: EntityTransactionalQueryContext) => Promise<T>
+    transactionScope: (queryContext: EntityTransactionalQueryContext) => Promise<T>,
+    transactionConfig?: TransactionConfig
   ): Promise<T> {
+    assert(
+      transactionConfig === undefined,
+      'should not pass transactionConfig to an already created transaction'
```
ide, line 157: Is this slightly more accurate?
```suggestion
      'Should not pass transactionConfig to a nested transaction'
```

sundeeppeswani: I think an isolation level can be set as long as no queries have been run in the transaction yet, but this is probably the safest way of doing this without adding extra checks.

sundeeppeswani (approved):

packages/entity/src/EntityQueryContext.ts
ide: That's accurate but the failure mode is dangerous. Specifically, Postgres just prints a warning. So you could have code that intends to set the isolation level to something strict (like wanting a serializability guarantee) and it could end up a no-op and you'd never know.

packages/entity/src/EntityQueryContextProvider.ts
wschurman: Because we need it to be nullable for `EntityTransactionalQueryContext.runInTransactionIfNotInTransactionAsync` (asserting empty object isn't as clear), it turns out it is cleaner to keep this optional. (tried the non-optional stuff though)

packages/entity/src/EntityQueryContext.ts
```diff
@@ -8,6 +8,18 @@ export type PreCommitCallback = (
   ...args: any
 ) => Promise<any>;
 
+export enum TransactionIsolationLevel {
+  READ_UNCOMMITTED,
+  READ_COMMITTED,
+  SNAPSHOT,
```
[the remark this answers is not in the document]
wschurman: These were taken from knex's isolationLevels (their well-typed enum)
https://github.com/knex/knex/blob/master/types/index.d.ts#L2300
https://github.com/knex/knex/blob/master/lib/execution/transaction.js#L26

I'll reduce these down to the postgres adapter compatible ones and will leave it up to future PRs to add more if necessary when new adapter dialects are added.

ide (approved):

packages/entity/src/EntityQueryContext.ts
sundeeppeswani: Yeah, my thought was to either keep track of a given `transactionConfig` in a transaction and allow it to be set as long as no other queries had been run, *or* check for any pgsql warnings/output. Both seem overkill here.

PR #208, opened 2023-05-25
[fix] Make inbound edge declaration asynchronous so import may be used

wschurman:
# Why

We're seeing import cycle issues again. It seems that not even putting these in a function is sufficient, and the only thing that works is making these use asynchronous imports in cyclic cases to delay even the importing (which is what is causing the cycle).

# How

For example, this will allow us to do the following:

```
getInboundEdges: async () => {
    return (
      await Promise.all([
        import('../EntityA'),
        import('../EntityB'),
      ])
    ).map((ec) => ec.default);
  },
```

This fixes the case where even just doing the import of `EntityA` at the top-level causes a dependency cycle at the module top-level.

# Test Plan

`yarn tsc`


quinlanj (approved):

ide:
If possible I'd like to separate using lazy imports for performance from using lazy imports for cycles. We might want this PR anyway but I'd like to understand why the cycles are causing a problem. Could you put up a branch I can take a look at tomorrow?

packages/entity/src/EntityConfiguration.ts
```diff
@@ -13,7 +13,7 @@ export default class EntityConfiguration<TFields> {
   readonly cacheableKeys: ReadonlySet<keyof TFields>;
   readonly cacheKeyVersion: number;
 
-  readonly getInboundEdges: () => IEntityClass<any, any, any, any, any, any>[];
+  readonly getInboundEdges: () => Promise<IEntityClass<any, any, any, any, any, any>[]>;
```
ide, line 16: API discussion topic: should we let developers write synchronous functions for getInboundEdges?

The advantage is that existing code that uses Entity doesn't need to be migrated. Also, direct synchronous calls produce cleaner stack traces and we can skip a tick of the async microtask queue.

The disadvantage is that there are now two ways to implement this method instead of just one. There's also a slight disadvantage in that the caller of this method might want to write: `if (typeof maybePromise.then === 'function') { await maybePromise }`, though it's also OK to write `await maybePromise` and non-promises will work fine.

```suggestion
  readonly getInboundEdges: () => IEntityClass<any, any, any, any, any, any>[] |
                                  Promise<IEntityClass<any, any, any, any, any, any>[]>;
```

wschurman: I think my preference is to keep types as strict as possible (and to have the types drive the runtime source of truth). In some repos I think something like this makes sense, but in this repo the types are as strict as possible.

As far as migrations go, large typescript migrations are pretty standard for this library as well.

ide (commented):
The code in this PR looks technically correct to me. Before approving I'd like to decide this is how we want to address cyclic initialization (I sent Will an in-depth writeup). There are some other ways we could handle cycles.

Also, I am slightly wary about making a configuration method be asynchronous. We just want to allow dynamic imports, which need to be asynchronous in JS, but we don't want external I/O to occur in getInboundEdges.

wschurman:
Superseded by https://github.com/expo/entity/pull/210.

PR #210, opened 2023-06-09
feat: entity companion definition static thunk

wschurman:
# Why

We're seeing an increasing number of circular reference reports caused by the requirement to maintain the companion definition singularity/singleton property in the application, usually by defining it in a module-top-level property. This has historically caused issues due to some properties in the companion/configuration using top-level definitions of entity types or other classes. 

In the past, we fixed this piece-by-piece, making the companion definition or configuration fields that could cause circular references functions so that circular references would be less likely. https://github.com/expo/entity/pull/89 https://github.com/expo/entity/pull/93

This PR expands upon this but should fix it once-and-for-all by maintaining the singleton requirement in the framework itself rather than pushing the responsibility to the application.

# How

This PR does a few things sequentially (yet procedurally) to accomplish the goal:
1. `getCompanionDefinition` -> `defineCompanionDefinition` - this signals to the application developer that this is a definition and that the framework will handle the singleton.
2. Instantiate a companion definition instance in `EntityCompanionProvider`. This will serve as the instance of the entity companion for the entity and is responsible for making this a singleton. Note that this doesn't need to be a singleton since these are objects, but since this method is called so frequently this memoization is a nice speedup.
3. Make all companion definition accesses occur through `EntityCompanionProvider`, for example: `viewerContext.entityCompanionProvider.getCompanionForEntity(EntityClass).entityCompanionDefinition.entityConfiguration.tableName`
4. Change thunk fields of `EntityCompanionDefinition` and `EntityConfiguration` to non-thunks due to the top-level being a thunk itself and accomplishing the lazy evaluation.
5. Move entity instantiation to `EntityLoader.constructEntity` to make it type-safe and prevent needing to access companion definition in the entity instances themselves (which would be infinite loop). This requires changing the constructor signature for `Entity` itself but that shouldn't be an issue since it is so rarely used outside of the framework itself.

# Test Plan

Run all tests.

Also `yarn link` this into a large codebase (Expo www) and ensure that this fixes circular issues and doesn't break anything.

### DIFF
--- a/packages/entity/src/Entity.ts
+++ b/packages/entity/src/Entity.ts
@@ -26,7 +26,7 @@ import ViewerContext from './ViewerContext';
  * own EntityCompanionDefinition.
  */
 export default abstract class Entity<
-  TFields,
+  TFields extends object,
   TID extends NonNullable<TFields[TSelectedFields]>,
   TViewerContext extends ViewerContext,
   TSelectedFields extends keyof TFields = keyof TFields
@@ -38,7 +38,7 @@ export default abstract class Entity<
    * @returns mutator for creating an entity
    */
   static creator<
-    TMFields,
+    TMFields extends object,
     TMID extends NonNullable<TMFields[TMSelectedFields]>,
     TMViewerContext extends ViewerContext,
     TMViewerContext2 extends TMViewerContext,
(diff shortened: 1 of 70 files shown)


packages/entity/src/Entity.ts
```diff
@@ -331,8 +331,13 @@ export interface IEntityClass<
...
 > {
-  new (viewerContext: TViewerContext, obj: Readonly<TFields>): TEntity;
-  getCompanionDefinition(): EntityCompanionDefinition<
+  new (constructorParam: {
+    viewerContext: TViewerContext;
+    id: TID;
+    databaseFields: Readonly<TFields>;
+    selectedFields: Readonly<Pick<TFields, TSelectedFields>>;
+  }): TEntity;
+  defineCompanionDefinition(): EntityCompanionDefinition<
```
ide, line 340: Consider documenting this method to explain that it is called once and memoized by the Entity framework.

ide (approved):

PR #211, opened 2023-06-09
chore: upgrade dependencies

wschurman:
# Why

This upgrades typescript and bumps minimum node version to 16.

This upgrades a good chunk of the dependencies:
- jest: tested by running all tests
- eslint: tested by running lint
- typedoc: tested by running `yarn typedoc`
- knex and pg: tested by running integration tests (note that pg 8.11.0 has a top-level return that is incompatible with our jest: https://github.com/brianc/node-postgres/issues/2980 so I locked it at 8.10.0 for now)
- example package demonstrating apollo, knex, etc: tested by running tests


tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
-    "target": "es2019",
-    "lib": ["es2019"],
+    "target": "es2021",
```
ide, line 3: Out of scope of this PR but take a look at presets like https://www.npmjs.com/package/@tsconfig/node16 that set some of these options for the target Node version.

ide (approved):

PR #215, opened 2023-11-22
feat: Add global mutation trigger field to EntityCompanionProvider

Josh-McFarlin:
# Why

Adding global mutation triggers to `EntityCompanionProvider` allows the creation of mutation triggers that run for every entity.

# How

I added the property `globalMutationTriggers` to `EntityCompanionProvider`. In the `EntityMutator` constructor, the `mutationTriggers` property is set to the combination of `companionProvider.globalMutationTriggers` and any mutation triggers included in the constructor's parameter. Global mutation triggers will always run before any mutation triggers defined for the individual entity.

# Test Plan

I modified [`EntityMutator-test.ts`](https://github.com/expo/entity/compare/main...global-mutation-triggers#diff-2eef1b2e4c6db76d391a6152d92d5b113a2429d99af476cc521c79f07590692a) to include the `globalMutationTriggers` during `createEntityMutatorFactory`, and these global triggers have their trigger count verified later.


packages/entity/src/EntityCompanionProvider.ts
```diff
@@ -139,14 +139,22 @@ export default class EntityCompanionProvider {
...
    */
   constructor(
     public readonly metricsAdapter: IEntityMetricsAdapter,
     private databaseAdapterFlavors: ReadonlyMap<
       DatabaseAdapterFlavor,
       DatabaseAdapterFlavorDefinition
     >,
-    private cacheAdapterFlavors: ReadonlyMap<CacheAdapterFlavor, CacheAdapterFlavorDefinition>
+    private cacheAdapterFlavors: ReadonlyMap<CacheAdapterFlavor, CacheAdapterFlavorDefinition>,
+    readonly globalMutationTriggers: EntityMutationTriggerConfiguration<
```
wschurman, line 151: This looks good. Definitely the right place for it, and a sensible default.

packages/entity/src/EntityMutator.ts
```diff
@@ -78,7 +78,35 @@ abstract class BaseMutator<
     >,
     protected readonly databaseAdapter: EntityDatabaseAdapter<TFields>,
     protected readonly metricsAdapter: IEntityMetricsAdapter
-  ) {}
+  ) {
+    this.mutationTriggers = [companionProvider.globalMutationTriggers, mutationTriggers].reduce<
```
wschurman, line 82: Instead of doing the merge here (which runs every time we create a mutator), I think we can do the merge earlier in the `EntityCompanion` constructor, which runs just once per entity type.

```diff
       this.privacyPolicy,
       entityCompanionDefinition.mutationValidators ?? [],
-      entityCompanionDefinition.mutationTriggers ?? {},
+      mergeEntityMutationTriggerConfigurations([
+        entityCompanionDefinition.mutationTriggers ?? {},
+        entityCompanionProvider.globalMutationTriggers ?? {},
+      ]),
       this.entityLoaderFactory,
       tableDataCoordinator.databaseAdapter,
```

Then, for the merging, I've found it to be a bit easier to read to explicitly list all the properties (which I agree is verbose, but it's hard to express types of keys of interfaces well in typescript. in this repo, having strict types is critical since it's primarily how we do breaking changes):

```typescript
function nonNullish<TValue>(value: TValue | null | undefined): value is NonNullable<TValue> {
  return value !== null && value !== undefined;
}

function mergeEntityMutationTriggerConfigurations<
  TFields extends object,
  TID extends NonNullable<TFields[TSelectedFields]>,
  TEntity extends ReadonlyEntity<TFields, TID, TViewerContext, TSelectedFields>,
  TSelectedFields extends keyof TFields,
>(
  mutationTriggerConfigurations: EntityMutationTriggerConfiguration<
    TFields,
    TID,
    TViewerContext,
    TEntity,
    TSelectedFields
  >[],
): EntityMutationTriggerConfiguration<TFields, TID, TViewerContext, TEntity, TSelectedFields> {
  return {
    beforeCreate: mutationTriggerConfigurations.flatMap((c) => c.beforeCreate).filter(nonNullish),
    afterCreate: mutationTriggerConfigurations.flatMap((c) => c.afterCreate).filter(nonNullish),
    beforeUpdate: mutationTriggerConfigurations.flatMap((c) => c.beforeUpdate).filter(nonNullish),
    afterUpdate: mutationTriggerConfigurations.flatMap((c) => c.afterUpdate).filter(nonNullish),
    beforeDelete: mutationTriggerConfigurations.flatMap((c) => c.beforeDelete).filter(nonNullish),
    afterDelete: mutationTriggerConfigurations.flatMap((c) => c.afterDelete).filter(nonNullish),
    beforeAll: mutationTriggerConfigurations.flatMap((c) => c.beforeAll).filter(nonNullish),
    afterAll: mutationTriggerConfigurations.flatMap((c) => c.afterAll).filter(nonNullish),
    afterCommit: mutationTriggerConfigurations.flatMap((c) => c.afterCommit).filter(nonNullish),
  };
}
```

packages/entity/src/__tests__/EntityMutator-test.ts
wschurman, line 1: And then assuming we do the merging in `EntityCompanion` instead of `EntityMutator`, the only test we need is a case in `EntityCompanion-test` to ensure that the resulting mutationTrigger array in the EntityMutatorFactory created in the constructor contains both local and global triggers. (this should hopefully be far easier to write than changing the `EntityMutator-test.ts`). You may need to access a private property:

```typescript
companion.getMutatorFactory()['mutationTriggers']
```

(we'd also probably want a unit test for `mergeMutationTriggerConfigurations`)

tsconfig.json
```diff
@@ -17,6 +17,7 @@
     "exactOptionalPropertyTypes": true,
     "noImplicitOverride": true,
     "esModuleInterop": true,
+    "skipLibCheck": true,
```
wschurman, line 20: revert

yarn.lock
wschurman, line 1: revert changes to this file. if running `yarn` in the root causes this to be dirty in the tree on your machine let me know and I can help debug

wschurman (changes requested):

packages/entity/src/EntityMutator.ts
Josh-McFarlin: Agreed!

packages/entity/src/utils/mergeMutationTriggers.ts
```diff
@@ -0,0 +1,44 @@
...
+    beforeDelete: mutationTriggerConfigurations.flatMap((c) => c.beforeDelete).filter(nonNullish),
+    afterDelete: mutationTriggerConfigurations.flatMap((c) => c.afterDelete).filter(nonNullish),
+    beforeAll: mutationTriggerConfigurations.flatMap((c) => c.beforeAll).filter(nonNullish),
+    afterAll: mutationTriggerConfigurations.flatMap((c) => c.afterAll).filter(nonNullish),
+    afterCommit: mutationTriggerConfigurations.flatMap((c) => c.afterCommit).filter(nonNullish),
+  };
+
+  /** Remove any trigger that is an empty array */
+  for (const key of Object.keys(merged) as (keyof typeof merged)[]) {
+    if (!merged[key]?.length) {
```
wschurman, line 38: since we can guarantee that all keys in merged have at least a value of `[]`:

```suggestion
    if (merged[key].length === 0) {
```

wschurman (approved):
Just a few changes before commit, otherwise looks good.

One other thing is I renamed the PR to use conventional commits style, so when you merge it ensure it keeps the new style in the final commit title. This just makes it so that the changelog is auto-generated.

PR #219, opened 2024-04-02
feat: require explicit query context specification

wschurman:
# Why

This is a RFC.

The proposal is to require the `queryContext` argument to reduce the likelihood of accidental omittance within a transaction block in application code.

At Expo, we've noticed sometimes a transaction will be started and the loader/mutator calls within it may accidentally not include the query context (which must be explicitly supplied as an arg to run within the transaction):
```typescript
await viewerContext.runInTransactionAsync(async (queryContext) => {
  const existingNote = await NoteEntity.loader(viewerContext) // note how queryContext was accidentally omitted but tsc didn't have an error
    .enforcing()
    .loadByIDAsync(args.id);
  return await NoteEntity.updater(existingNote, queryContext)
    .setField('title', args.note.title)
    .setField('body', args.note.body)
    .enforceUpdateAsync();
});
```

# How

The proposed change is to always require `queryContext`. At Expo, we added a method to our `ViewerContext` subclass to vend the default `EntityQueryContext` (non-transactional) for our default database adaptor flavor. I added a similar method to `TestViewerContext` within this repo to demonstrate the convenience accessor.

# RFC

- Does this pattern make sense?
- Is the default query context convenience pattern appropriate?

# Test Plan

`yarn tsc`

### DIFF
--- a/packages/entity-cache-adapter-local-memory/src/__tests__/GenericLocalMemoryCacher-full-test.ts
+++ b/packages/entity-cache-adapter-local-memory/src/__tests__/GenericLocalMemoryCacher-full-test.ts
@@ -25,13 +25,19 @@ describe(GenericLocalMemoryCacher, () => {
     const cacheKeyMaker = genericCacher['makeCacheKey'].bind(genericCacher);
 
     const date = new Date();
-    const entity1Created = await LocalMemoryTestEntity.creator(viewerContext)
+    const entity1Created = await LocalMemoryTestEntity.creator(
+      viewerContext,
+      viewerContext.getQueryContextForDatabaseAdaptorFlavor('postgres')
+    )
       .setField('name', 'blah')
       .setField('dateField', date)
       .enforceCreateAsync();
 
     // loading an entity should put it in cache
-    const entity1 = await LocalMemoryTestEntity.loader(viewerContext)
+    const entity1 = await LocalMemoryTestEntity.loader(
+      viewerContext,
+      viewerContext.getQueryContextForDatabaseAdaptorFlavor('postgres')
+    )
       .enforcing()
       .loadByIDAsync(entity1Created.getID());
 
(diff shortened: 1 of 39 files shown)


ide (commented):
I'm open to this and think it probably is good to make the query context explicit.

How independent are VCs and QCs? Should VCs no longer have a `getQueryContextForDatabaseAdaptorFlavor()` method if application code that works with VCs will always have a QC anyway since VCs and QCs are now passed around together?

What are the potential consequences if a VC is passed in with a QC, but `vc.getQueryContextForDatabaseAdaptorFlavor() !== qc`?

Should we pair together VCs and QCs in an object like: `interface EntityContext { viewerContext; queryContext; }`? (I'm thinking probably no for now. Keep the values flatter.)

quinlanj (approved):
I'd be very happy with something like this in www as I'm always super paranoid of forgetting to pass a qc.

wschurman:
> How independent are VCs and QCs? Should VCs no longer have a `getQueryContextForDatabaseAdaptorFlavor()` method if application code that works with VCs will always have a QC anyway since VCs and QCs are now passed around together?

They are related in that a ViewerContext is the entry point into the entity framework from an application (it has the reference to the EntityCompanionProvider which is the "instance" of the entity framework). So for an application to get a vended QueryContext, it must go through the viewer context by definition. But they are not tied together and are independent. There can be many flavors of viewer context (depending on the application) and both transactional and non-transactional query contexts being used for any loader/mutator call at any point.

> What are the potential consequences if a VC is passed in with a QC, but `vc.getQueryContextForDatabaseAdaptorFlavor() !== qc`?

I think maybe we should rename `ViewerContext.getQueryContextForDatabaseAdaptorFlavor` to `ViewerContext.getNonTransactionalQueryContextForDatabaseAdaptorFlavor` or something. In the `PostgresEntityQueryContextProvider`, this is just a reference to the global (singleton) knex instance.

So, to answer your question, all non-transactional query context (`EntityNonTransactionalQueryContext`) are equivalent in that they just forward the calls to the knex singleton (for postgres). TransactionalQueryContext though are unique and reference a single transaction; they're vended via `runInTransactionAsync((vendedQueryContext) => ...`

> Should we pair together VCs and QCs in an object like: `interface EntityContext { viewerContext; queryContext; }`? (I'm thinking probably no for now. Keep the values flatter.)

I'd say no for now, but if/when we have more database adaptor flavors we may need to reassess. But even then, one could imagine just having single instance of a viewer context being passed around and then multiple query contexts being used for multiple databases. This is all theoretical though.

ide (approved):

wschurman:
Reverted due to the integration of this into our first-party codebase (www) being too unwieldy (6k tsc errors). Going to come up with a different approach, maybe something more codemod-able.

PR #221, opened 2024-04-08
feat: Allow UUIDv6/7/8 in UUIDField validation

ide:
Why
---
Currently the UUIDField type supports only UUIDv1-5 and to use a newer UUID version we need to use StringField.

How
---
Manually implemented UUID validation, adding support for versions 6, 7, and 8. Also added support for the "max" UUID, which is now part of the spec.

Test Plan
---
Added a UUIDv7 value (from https://uuid7.com/) to the test cases.

### DIFF
--- a/packages/entity/src/EntityFields.ts
+++ b/packages/entity/src/EntityFields.ts
@@ -1,6 +1,8 @@
-import { validate as validateUUID } from 'uuid';
-
 import { EntityFieldDefinition } from './EntityFieldDefinition';
+
+// Use our own regex since the `uuid` package doesn't support validating UUIDv6/7/8 yet
+const UUID_REGEX =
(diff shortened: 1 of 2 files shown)


packages/entity/src/EntityFields.ts
```diff
@@ -1,7 +1,9 @@
-import { validate as validateUUID } from 'uuid';
```
wschurman, line 1: Should we make this a dev dependency?

wschurman (approved):

packages/entity/src/EntityFields.ts
ide: It turns out `uuid` is used by `StubDatabaseAdapter`, which ends up getting exported as part of the package's production interface. So for now the right thing to do AFAIK is to keep it as as dependency. If we introduced `@expo/entity-testing` as a separate package, we could make uuid a devDep of `@expo/entity` but I don't think that'd get us much.

PR #223, opened 2024-05-06
feat: add privacy policy rule reordering

wschurman:
# Why

This adds the ability to direct privacy policies to do "smart reordering" of the rules within a policy. The benefit here is that we can optimize the policy at evaluation time to do less-computationally-expensive rules earlier in the hopes that they are definitive in their evaluation results and therefore we don't evaluate the later more-computationally-expensive rules.

This comes with some notable caveats though so it is an opt-in behavior. The caveats are:
- Sometimes SKIP rules are used to do things like log, dry-run, etc. In this case, if they are reordered in an unexpected way, they may not log or execute.
- Sometimes rule execution order matters. Example:
    - AllowRule1 - Allow rule that checks if X is true and allows if so, otherwise skips
    - AllowRule2 - Allow rule that throws a hard exception if X is false, but always Allows.
    - In this instance, a non-reorderable policy with [AllowRule1, AllowRule2] would never throw a hard exception.
    - But if allowed to be reorderable, AllowRule2 may run before AllowRule1 (depending on complexity).
    - It is best not to write non-reorderable rules.

# How

1. Change privacy rule subclasses to make them categorize themselves into AllowOrSkip, Skip, or DenyOrSkip. (AllowOrDeny is not allowed).
2. Add complexity field to privacy policy rules. This is programmer-defined. Getting it wrong isn't the end of the world since it just means that the policy execution will be slightly more expensive at worst (not incorrect).
1. Add rule ordering and tests for the functionality.

# Test Plan

Run tests.

### DIFF
--- a/packages/entity/src/rules/PrivacyPolicyRuleEnums.ts
+++ b/packages/entity/src/rules/PrivacyPolicyRuleEnums.ts
@@ -19,23 +19,25 @@ export enum RuleEvaluationResult {
 }
 
 /**
- * Estimated complexity of a privacy policy rule.
+ * Estimated complexity of a privacy policy rule. Definition of what rules fall into what complexity category
+ * is application-specific, but the AlwaysAllow/AlwaysDeny/AlwaysSkip rules provided are all defined as LOW
+ * complexity.
  */
 export enum RuleComplexity {
   /**
-   * Rule doesn't make any async calls.
+   * Rule complexity is low. Definition dependent on application.
    */
-  CONSTANT_TIME = 0,
+  LOW = 0,
 
   /**
-   * Rule makes an small amount of async calls. Definition of small dependent on application.
+   * Rule complexity is medium. Definition dependent on application.
    */
-  SMALL_ASYNC = 1,
+  MEDIUM = 1,
 
   /**
-   * Rule makes a large amount of async calls. Definition of large dependent on application.
+   * Rule complexity is high. Definition dependent on application.
    */
-  LARGE_ASYNC = 2,
+  HIGH = 2,
 }
 
 /**
--- a/packages/entity/src/rules/PrivacyPolicyRuleEnums.ts
+++ b/packages/entity/src/rules/PrivacyPolicyRuleEnums.ts
@@ -1,0 +1,59 @@
+/**
+ * Enum return type of privacy policy rule evaluation.
+ */
+export enum RuleEvaluationResult {
+  /**
+   * Deny viewer access to the entity.
+   */
+  DENY = -1,
+
+  /**
+   * Defer entity viewer access to subsequent rule in the privacy policy.
+   */
+  SKIP = 0,
+
+  /**
+   * Allow viewer access to the entity.
+   */
+  ALLOW = 1,
+}
+
+/**
+ * Estimated complexity of a privacy policy rule.
+ */
+export enum RuleComplexity {
+  /**
+   * Rule doesn't make any async calls.
+   */
+  CONSTANT_TIME = 0,
+
+  /**
+   * Rule makes an small amount of async calls. Definition of small dependent on application.
+   */
+  SMALL_ASYNC = 1,
+
+  /**
+   * Rule makes a large amount of async calls. Definition of large dependent on application.
+   */
+  LARGE_ASYNC = 2,
+}
+
+/**
+ * Used for disambiguation between classes of privacy policy rules at runtime.
+ * Multiple rules of the same type in a row can be reordered safely to improve policy evaluation time
+ * based on complexities.
+ */
+export enum RuleEvaluationResultType {
+  /**
+   * Rule is a DenyOrSkipPrivacyPolicyRule. Can be safely reordered with other sequential DenyOrSkipPrivacyPolicyRule or SkipPrivacyPolicyRule in a policy.
+   */
+  DENY_OR_SKIP,
+  /**
+   * Rule is a SkipPrivacyPolicyRule. Can be safely reordered with any other sequential privacy policy rules in a policy.
+   */
+  SKIP,
+  /**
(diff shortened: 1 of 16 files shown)


packages/entity/src/rules/PrivacyPolicyRuleEnums.ts
```diff
@@ -0,0 +1,59 @@
...
+  /**
+   * Rule makes an small amount of async calls. Definition of small dependent on application.
+   */
+  SMALL_ASYNC = 1,
+
+  /**
+   * Rule makes a large amount of async calls. Definition of large dependent on application.
+   */
+  LARGE_ASYNC = 2,
+}
```
ide, line 39: These enum value names mix several types of complexity:

- Async vs. sync
- I/O vs. no I/O
- Run time complexity - O(1), O(n), ...

An alternative way to categorize the complexity would be: NO_DATA_FETCHING, SMALL_DATA_FETCHING, LARGE_DATA_FETCHING.

Either way I think it would be useful to provide more guidance - even if it's just internally - about small vs. large.

For example: If you need to fetch an associated entity of an associated entity, that's just two entities but two round trips. If you need to fetch n associated entities, that's just one round trip. How do you compare these two access patterns?

packages/entity/src/rules/PrivacyPolicyRuleEnums.ts
```diff
@@ -0,0 +1,59 @@
...
+  DENY_OR_SKIP,
+  /**
+   * Rule is a SkipPrivacyPolicyRule. Can be safely reordered with any other sequential privacy policy rules in a policy.
+   */
+  SKIP,
+  /**
+   * Rule is an AllowOrSkipPrivacyPolicyRule. Can be safely reordered with other sequential AllowOrSkipPrivacyPolicyRule or SkipPrivacyPolicyRule in a policy.
+   */
+  ALLOW_OR_SKIP,
+}
```
ide, line 59: While I think it's fine to have this enum and it's not hard to maintain, wouldn't `instanceof` checks achieve the same thing so that the enums and abstract classes don't need to be kept in sync? (But in practice people aren't writing new abstract rule classes.)

wschurman: Yep, abstract classes work as well. I've been tending to like having fields for simpler discriminated unions (since using `instanceof` checks in a switch is somewhat clunky in typescript, though better in kotlin). In this instance, both work.

packages/entity/src/rules/PrivacyPolicyRuleEnums.ts
wschurman: Okay, switching to just a number didn't go as well since there are default rules provided by the entity package that need a complexity rating.

Instead, I just made the categories LOW/MEDIUM/HIGH and left categorization and definition up to the application.

packages/entity/src/rules/PrivacyPolicyRuleEnums.ts
```diff
@@ -0,0 +1,61 @@
...
+   */
+  ALLOW = 1,
+}
+
+/**
+ * Estimated complexity of a privacy policy rule. Definition of what rules fall into what complexity category
+ * is application-specific, but the AlwaysAllow/AlwaysDeny/AlwaysSkip rules provided are all defined as LOW
+ * complexity.
+ */
+export enum RuleComplexity {
```
ide, line 26: Last suggestion is to call this cost instead of complexity since complexity tends to imply O(?) but a privacy policy could be costly if it does a waterfall of data fetches even if it runs in O(1).

ide (approved):

packages/entity/src/rules/internal/PrivacyPolicyRuleOrdering.ts
```diff
@@ -0,0 +1,70 @@
...
+ * Privacy policy list of PrivacyPolicyRule can be reordered safely according to
+ * the following constraints on RuleEvaluationResultType:
+ * 1. Two adjacent ALLOW_OR_SKIP rules can be swapped.
+ * 2. Two adjacent DENY_OR_SKIP rules can be swapped.
+ * 3. SKIP rules can be swapped with any other adjacent rule.
+ *
+ * @param rulesetOriginalOrder - ruleset with original ordering defined in code
+ * @returns reordered ruleset depending on constraints above
+ */
+export function reorderRulesByRuleComplexityGroups<
```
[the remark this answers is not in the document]
wschurman: That could definitely work, too. Here's some thoughts:
- I think the main issue I see is that we don't have a way of enforcing that an actual unit test is run (that'd live in the application). That being said, we could just do the validation at evaluation time (validate that it is in the optimal order) and throw an invariant error if it isn't (not a runtime error). I somewhat like this idea. I also like the easier-to-debug nature of it.
- Alternatively, going off that idea, this could live entirely in the application layer and not be in the framework at all. Theoretically we could just have a lint rule to require all privacy rules have a complexity field and then have a second lint rule that analyzes their ordering.
- The natural extension of this is to calculate complexity at runtime, yep. Or even use some other heuristic to order them. In other similar frameworks, I've seen it even reordered according to which rule is most-often the deciding rule.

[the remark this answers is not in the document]
quinlanj: Summarizing our discussion slack:
- For an application that would only ever hard code the complexity, I like the lint rule idea. It's unobtrusive and allows for easier debugging at runtime
- In the long term if we wanted to make a dynamic `complexity` system (ie) optional concept/field called “PrivacyPolicyOrderer” that had the ability to reorder, used postgres stats to update complexity every now and then, what you have already is probably fine

quinlanj (approved):
approving to unblock :)

wschurman:
Going to look into the suggestion about just doing the ordering validation via a lint rule. Will re-open if we decide to go with this.

PR #232, opened 2024-05-31
feat: add previousValue to privacy policy context for updates

wschurman:
# Why

One authorization evaluation case wasn't considered when building this context: when we want to allow a user to make a final update to an entity based on their permission from the value of the entity before the update, and then not allow them to make any further updates.

This PR adds previousValue to context for updates, somewhat like what we do for validators and triggers.

Closes ENG-12402.

# Test Plan

Run tests (there's a test that looks for this value).


packages/entity/src/EntityPrivacyPolicy.ts
```diff
@@ -11,7 +11,19 @@ import PrivacyPolicyRule, { RuleEvaluationResult } from './rules/PrivacyPolicyRu
...
+  TFields extends object,
+  TID extends NonNullable<TFields[TSelectedFields]>,
+  TViewerContext extends ViewerContext,
+  TEntity extends ReadonlyEntity<TFields, TID, TViewerContext, TSelectedFields>,
+  TSelectedFields extends keyof TFields = keyof TFields
+> = {
+  /**
+   * When this privacy policy is being evaluated as a result of an update, this will be populated with the value
+   * of the entity before the update. Note that this doesn't only apply to UPDATE authorization actions though:
+   * when an entity is updated it is re-loaded after the update completes.
```
ide, line 24: Would it be clearer to say:
```suggestion
   * when an entity is updated it is re-LOADed after the update completes.
```
to highlight why the LOAD is different from the UPDATE?

ide (approved):

kitten (approved):
Can't say I know the codebase well enough to give a proper review ❤️ But the change looks great, and having this capability is awesome! Code overall makes sense to me

PR #233, opened 2024-06-03
fix: always reload entity after update since cascading changes may have changed it since commit

wschurman:
# Why

Noticed this while reading through mutator code. The issue is that previously we would not re-load the entity if `skipDatabaseUpdate` was true since we assumed the entity was unchanged. But there are some scenarios where this isn't true, namely `SET_NULL_INVALIDATE_CACHE_ONLY`.

For example, if a field in this entity is set null when cascade deleting, we want to reload it before calling the after-update triggers (especially the post-commit ones).

Closes ENG-12407.

# How

Always re-load the entity. Note that this requires an explicit check to ensure a stable ID. Previously this check was implicitly enforced by the database adapter.

# Test Plan

Run new tests.

### DIFF
--- a/packages/entity/src/EntityMutator.ts
+++ b/packages/entity/src/EntityMutator.ts
@@ -422,6 +422,7 @@ export class UpdateMutator<
     cascadingDeleteCause: EntityCascadingDeletionInfo | null
   ): Promise<Result<TEntity>> {
     this.validateFields(this.updatedFields);
+    this.ensureStableIDField(this.updatedFields);
 
     const entityLoader = this.entityLoaderFactory.forLoad(this.viewerContext, queryContext, {
       previousValue: this.originalEntity,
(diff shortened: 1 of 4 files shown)


Josh-McFarlin (approved):
lgtm

packages/entity/src/EntityMutator.ts
```diff
@@ -517,6 +514,14 @@ export class UpdateMutator<
 
     return result(updatedEntity);
   }
+
+  private ensureStableIDField(updatedFields: Partial<TFields>): void {
+    const originalId = this.originalEntity.getID();
+    const idField = this.entityConfiguration.idField;
+    if (idField in updatedFields && originalId !== updatedFields[idField]) {
```
ide, line 521: In practice this won't matter, but if idField were "constructor" or "toString", we don't want to confuse those with Object.prototype.constructor, etc.

```suggestion
    if (updatedFields.hasOwnProperty(idField) && originalId !== updatedFields[idField]) {
```

It probably wouldn't be a bad idea to ban fields named any of the following:
```
Object.prototype.__proto__
Object.prototype.constructor
Object.prototype.hasOwnProperty
Object.prototype.isPrototypeOf
Object.prototype.propertyIsEnumerable
Object.prototype.toLocaleString
Object.prototype.toString
Object.prototype.valueOf
```

Technically it is possible to write code in a way that allows for these fields (use `hasOwnProperty()` instead of `in`, including `for ... in`) but "constructor" is the only field I could see someone wanting to use and they should just choose another name and move on.

ide (approved):

packages/entity/src/EntityMutator.ts
wschurman: Will do the second portion of disallowing prototype names as field names in a follow-up PR.

wschurman: https://github.com/expo/entity/pull/236

PR #235, opened 2024-06-04
fix: constrain entity fields type to string-keyed object

wschurman:
# Why

This allows us to remove the `any` typecast in EntityConfiguration.ts and is more correct: only string-keyed field objects are allowed.

This is in preparation for addressing this comment: https://github.com/expo/entity/pull/233#discussion_r1625121092

# How

Add type constraint where necessary.

# Test Plan

`yarn tsc`


Josh-McFarlin:
Is the plan to replace `T extends Record<string, any>` with a type omitting prototype names?

wschurman:
> Is the plan to replace `T extends Record<string, any>` with a type omitting prototype names?

Nope, I just noticed the `any` cast at the callsite where I was planning to add validation and decided that it'd be easy enough to get rid of the any cast (which is a code smell).

That being said, your idea is interesting and I wonder if it's possible. I think the typescript Record type extends Object which has those properties.

Josh-McFarlin (approved):
lgtm

wschurman:
It seems like omitting Object core types is not supported: https://www.typescriptlang.org/play/?#code/GYVwdgxgLglg9mABAIwDwBUBiMCmAbAEwGdEcAPKHMYxAeQFsYpUAlHCOAJwNSKk5hgA5gBpEAQzABPAHxiA5FDgBlfoKHyZMgBTBchIgC5EWfcQCUxgG5wYBRAG8AUIleI9+YgDolqgcO1zAG4nAF8gA

ide (approved):

PR #236, opened 2024-06-04
fix: disallow keys of JS Object prototype for safety

wschurman:
# Why

This PR addresses this comment: https://github.com/expo/entity/pull/233#discussion_r1625121092

Essentially, we want to disallow naming entity fields the same as some predefined JS prototype fields so that code in entity that uses `for .. in` loops over the fields object aren't accidentally unsafe (it's too easy a mistake to make).

# How

Disallow fields, add test.

# Test Plan

Run test.


packages/entity/src/EntityConfiguration.ts
```diff
@@ -85,6 +85,20 @@ export default class EntityConfiguration<TFields extends Record<string, any>> {
     this.dbToEntityFieldsKeyMapping = invertMap(this.entityToDBFieldsKeyMapping);
   }
 
+  private static validateSchema<TFields extends Record<string, any>>(schema: TFields): void {
+    const disallowedFieldsKeys = Object.getOwnPropertyNames(Object.getPrototypeOf({}));
```
ide, line 89:
```suggestion
    const disallowedFieldsKeys = Object.getOwnPropertyNames(Object.prototype);
```

See:
```
> Object.getPrototypeOf({}) === Object.prototype
true
```

packages/entity/src/EntityConfiguration.ts
```diff
@@ -85,6 +85,20 @@ export default class EntityConfiguration<TFields extends Record<string, any>> {
...
+    for (const disallowedFieldsKey of disallowedFieldsKeys) {
+      // when `hasOwnProperty` is a field name, we can't call it as a method. it's still invalid though.
+      if (typeof schema.hasOwnProperty !== 'function') {
+        throw new Error(`Entity field name not allowed: hasOwnProperty`);
+      }
```
Josh-McFarlin, line 94: Can this be moved out of the loop?

packages/entity/src/EntityConfiguration.ts
```diff
@@ -85,6 +85,20 @@ export default class EntityConfiguration<TFields extends Record<string, any>> {
...
+
+      if (schema.hasOwnProperty(disallowedFieldsKey)) {
+        throw new Error(`Entity field name not allowed: ${disallowedFieldsKey}`);
+      }
```
ide, line 98: 1. Good thinking to test for `hasOwnProperty` specially, but `Object.hasOwn()` takes care of this :) (Before `Object.hasOwn` was added, people would write `Object.prototype.hasOwnProperty.call(object, fieldName)` to handle the issue you identified.)
2. A more detailed error message helps specify "why" the name is disallowed.

```suggestion
      if (Object.hasOwn(schema, disallowedFieldsKey)) {
        throw new Error(`Entity field name not allowed to prevent conflicts with standard Object prototype fields: ${disallowedFieldsKey}`);
      }
```

ide (changes requested):

packages/entity/src/EntityConfiguration.ts
```diff
@@ -7,7 +7,7 @@ import { mapMap, invertMap, reduceMap } from './utils/collections/maps';
  * The data storage configuration for a type of Entity. Contains information relating to IDs,
  * cachable fields, field mappings, and types of cache and database adapter.
  */
-export default class EntityConfiguration<TFields extends Record<string, any>> {
+export default class EntityConfiguration<TFields extends Omit<Record<string, any>, 'constuctor'>> {
```
ide, line 10:
```suggestion
export default class EntityConfiguration<TFields extends Omit<Record<string, any>, 'constructor'>> {
```

ide (approved):
Important typo to fix mentioned inline.

packages/entity/src/EntityConfiguration.ts
wschurman: Lol nice catch. This was me testing https://github.com/expo/entity/pull/235#issuecomment-2148216997. Will remove.

PR #239, opened 2024-06-05
chore: move entity loader utils into their own object

wschurman:
# Why

This is the second part of https://github.com/expo/entity/pull/238. From the summary:

> In a follow-up, I'll look into moving the common methods (the ones in knownLoaderOnlyDifferences) to a different utils class since it no longer makes sense for them not to be on the loader class. But to limit the scope of this PR, they are kept in the NonEnforcingEntityLoader class for now.

This PR is this follow-up. It creates a `utils` concept that the loaders use but also that end-users can use in rare cases (we use them in rare cases in Expo server code).

The reason for this is that it doesn't make sense to have a call of the form (what they are after #238):
`AssetEntity.loader(viewerContext, queryContext).nonEnforcing().invalidateEntityAsync(asset)`

Instead, this PR makes it:
`AssetEntity.loader(viewerContext, queryContext).utils().invalidateEntityAsync(asset)`

Before both PRs, it was:
`AssetEntity.loader(viewerContext, queryContext).invalidateEntityAsync(asset)`

But I think it's better to not have them on the top-level loader for cleanliness.

# Test Plan

Run all tests.

### DIFF
--- a/packages/entity-cache-adapter-local-memory/src/__tests__/GenericLocalMemoryCacher-full-test.ts
+++ b/packages/entity-cache-adapter-local-memory/src/__tests__/GenericLocalMemoryCacher-full-test.ts
@@ -83,7 +83,7 @@ describe(GenericLocalMemoryCacher, () => {
 
     // invalidate from cache to ensure it invalidates correctly
     await LocalMemoryTestEntity.loader(viewerContext)
-      .withAuthorizationResults()
+      .utils()
       .invalidateFieldsAsync(entity1.getAllFields());
     const cachedResultMiss = await entitySpecificGenericCacher.loadManyAsync([
       cacheKeyMaker('id', entity1.getID()),
--- a/packages/entity-cache-adapter-redis/src/__integration-tests__/BatchedRedisCacheAdapter-integration-test.ts
+++ b/packages/entity-cache-adapter-redis/src/__integration-tests__/BatchedRedisCacheAdapter-integration-test.ts
@@ -160,7 +160,7 @@ describe(GenericRedisCacher, () => {
 
     // invalidate from cache to ensure it invalidates correctly in both caches
     await RedisTestEntity.loader(viewerContext)
-      .withAuthorizationResults()
+      .utils()
       .invalidateFieldsAsync(entity1.getAllFields());
     await expect(redis.get(cacheKeyEntity1)).resolves.toBeNull();
     await expect(redis.get(cacheKeyEntity1NameField)).resolves.toBeNull();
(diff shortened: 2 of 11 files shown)


ide (changes requested):
From an API standpoint what makes you not want these methods on the top-level loader? When I think about the desired end state, loaders would be enforcing by default and most code would look like `XEntity.loader(vc, qc).operation()`. `XEntity.loader(vc, qc).nonEnforcing().operation()` would be the exception.

In terms of types, `XEntity.loader(vc, qc)` would conceptually return `EnforcingLoader | Utils` and `XEntity.loader(vc, qc).nonEnforcing()` would return just `NonenforcingLoader`.

quinlanj:
I prefer `AssetEntity.loader(viewerContext, queryContext).nonEnforcing().invalidateEntityAsync(asset)` over `AssetEntity.loader(viewerContext, queryContext).utils().invalidateEntityAsync(asset)` because `utils` gives me no information about what is going on.

Most of the calls in www are of form `AssetEntity.loader(viewerContext, queryContext).enforcing().invalidateEntityAsync(asset)` anyways, so the user has to be aware of the `Result` concept and there is no reason to abstract it away with `utils()`.

wschurman:
I think the reason I prefer them not be top-level is that they are so rarely used compared to loader methods, and shouldn't be used by normal applications except in exceptional cases. Some data from Expo backend:
- ~1200 instances of loader methods
- ~3 instances of `invalidateEntityAsync` calls
- 2 instances of `invalidateFieldsAsync`, both in tests

So to me the API that makes the most sense in the end (for now, we can't put enforcing loader methods on top-level yet due to implicit breaking change) is:
- Top-level loader has just three methods: `enforcing`, `withAuthorizationResults`, and `utils` (can be named something else)
- Each of those loaders only has `loadX` methods (no `invalidateEntityAsync`).

ide (approved):

wschurman:
> I prefer `AssetEntity.loader(viewerContext, queryContext).nonEnforcing().invalidateEntityAsync(asset)` over `AssetEntity.loader(viewerContext, queryContext).utils().invalidateEntityAsync(asset)` because `utils` gives me no information about what is going on.

This is a good point. "utils" in general is a pretty poor term. But in this case it's a bunch of utility methods that are both used by the loaders themselves as well as (super rarely) used by the application. They don't have a "enforcing" or "withAuthorizationResult" property (the methods in utils don't use results or throw errors or load entities), so in this case I think they belong elsewhere. And "utils" is all I got as far as creativity lol.

> Most of the calls in www are of form `AssetEntity.loader(viewerContext, queryContext).enforcing()` anyways, so the user has to be aware of the `Result` concept and there is no reason to abstract it away with `utils()`.

I think theres a slight misunderstanding in the set of methods being put in utils. It's just things like `invalidateEntityAsync`, but none of the load/result-based-load methods, those still live in their loaders.

So after this PR it is:
- `Entity.loader(vc, qc).enforcing().loadX`:  enforcing loader methods
- `Entity.loader(vc, qc).withAuthorizationResults().loadX`: non-enforcing loader methods (wraps authorization and construction errors in Result)
- `Entity.loader(vc, qc).utils().invalidateEntityAsync`: other methods related to loaders but very rarely used outside of loaders.

PR #243, opened 2024-06-24
feat: add EntityEdgeDeletionAuthorizationInferenceBehavior for canViewerDeleteAsync

wschurman:
# Why

This PR adds a new concept for optimized evaluation of `canViewerDeleteAsync`: `EntityEdgeDeletionAuthorizationInferenceBehavior`.

The way it works is: when the method goes to recursively authorize all cascading deletes/set-nulls for an entity being checked in `canViewerDeleteAsync`, an application developer can specify that checking one instance of an edge of an association type is sufficient for inferring that all edges share the same privacy checks and loading all of them isn't necessary. Without this, all edges are loaded and authorization to cascade delete / set null are checked individually.

This is more of a best-effort optimistic evaluation to tell ahead of time whether a deletion will succeed or fail, and is not used during actual deletion since an application developer may apply it incorrectly.

It was inspired by this comment: https://github.com/expo/entity/pull/224#pullrequestreview-2090051711

# How

Add concept with explanation in comments, add tests.

# Test Plan

Run tests.

### DIFF
--- a/packages/entity/src/EntityFieldDefinition.ts
+++ b/packages/entity/src/EntityFieldDefinition.ts
@@ -36,6 +36,35 @@ export enum EntityEdgeDeletionBehavior {
   SET_NULL,
 }
 
+export enum EntityEdgeDeletionAuthorizationInferenceBehavior {
+  /**
+   * Authorization to delete (when CASCADE_DELETE_INVALIDATE_CACHE_ONLY or CASCADE_DELETE) or update
+   * (when SET_NULL_INVALIDATE_CACHE_ONLY or SET_NULL) all entities at the ends of edges of this type
+   * cannot be inferred from authorization of any single entity at the end of an edge of this type.
+   *
+   * To evaluate canViewerDeleteAsync for the source entity, canViewerDeleteAsync must be called on all
+   * entities at the ends of all edges of this type.
+   */
+  NONE,
+
+  /**
+   * Authorization to delete (when CASCADE_DELETE_INVALIDATE_CACHE_ONLY or CASCADE_DELETE) or update
+   * (when SET_NULL_INVALIDATE_CACHE_ONLY or SET_NULL) all entities at the ends of edges of this type
+   * may be inferred from authorization of any single entity at the end of an edge of this type.
+   *
+   * To evaluate canViewerDeleteAsync for the source entity, canViewerDeleteAsync must only be called on
+   * a single entity at the end of one edge of this type chosen at random.
+   *
+   * This should only be the case when the entity at the other end of this edge can be implicitly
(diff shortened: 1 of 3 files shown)


wschurman:
* **#243** <a href="https://app.graphite.dev/github/pr/expo/entity/243?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @wschurman and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

wschurman:
One concept I'd like to be able to express alongside this (that I haven't figured out yet) is the ability to ensure `canViewerDeleteAsync` doesn't load too many entities unexpectedly due to a forgotten `edgeDeletionAuthorizationInferenceBehavior` specification somewhere in the cascade tree.

Put another way, the most common use of this will likely be doing a simple optimistic check ahead of long-running bulk/batch deletion of a tree of entities. It'd be nice to be able to limit how many entities can be loaded in that execution of `canViewerDeleteAsync` to prevent OOM.

packages/entity/src/EntityFieldDefinition.ts
```diff
@@ -93,6 +122,14 @@ export interface EntityAssociationDefinition<
...
    */
   edgeDeletionBehavior: EntityEdgeDeletionBehavior;
+
+  /**
+   * Optimization setting for evaluation of this edge in canViewerDeleteAsync. Can be used to optimize permission checks
+   * for edge cascading deletions based on application-specific design of the entity association. Not used during actual deletions.
+   *
+   * Defaults to EntityEdgeDeletionAuthorizationInferenceBehavior.NONE.
+   */
+  edgeDeletionAuthorizationInferenceBehavior?: EntityEdgeDeletionAuthorizationInferenceBehavior;
```
wschurman, line 132: Should this be required?
- Pros: it'd be harder for application developers to forget to add it when applicable.
- Cons: it's somewhat hard to understand as a concept, would require adding to all associations. Unnecessary for applications not using `canViewerDeleteAsync`.

```suggestion
  edgeDeletionAuthorizationInferenceBehavior: EntityEdgeDeletionAuthorizationInferenceBehavior;
```

ide: My gut feeling is to make it optional since it is an advanced concept and an optimization that makes sense only for large fanouts. In contrast, someone adding a 1:1 edge wouldn't need to implement this at all.

ide (approved):

PR #248, opened 2024-09-26
feat: allow cascadingDeleteCause to be passed in via public deletion mutator methods

wschurman:
# Why

When doing large batch deletions like the ones specified in https://github.com/expo/entity/pull/243, the case is usually to serially do a bunch of deletions ahead of a cascade in order to decrease memory usage.

This has one issue: these become normal deletions. Even though they're somewhat "part" of this cascade deletion.

Some privacy policies make use of the `cascadingDeleteCause` in order to determine delete-ability. Because of this, these "cascading" deletes should have the cause.

# How

Add the ability to manually specify it.

# Test Plan

N/A

### DIFF
--- a/packages/entity/src/Entity.ts
+++ b/packages/entity/src/Entity.ts
@@ -153,7 +153,7 @@ export default abstract class Entity<
       .getViewerScopedEntityCompanionForClass(this)
       .getMutatorFactory()
       .forDelete(existingEntity, queryContext)
-      .deleteAsync();
+      .deleteAsync(null);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


wschurman:
* **#248** <a href="https://app.graphite.dev/github/pr/expo/entity/248?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @wschurman and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

wschurman:
Closing since it's likely this library should just provide a batch deletion abstraction instead and inject it internally there. This field is meant to be constructed recursively only by the deletion mutator.

PR #255, opened 2025-02-20
chore: switch barreling to ctix

wschurman:
# Why

We use barrels in the exported modules to make imports nicer. [`barrelsby`](https://github.com/bencoveney/barrelsby) was the best option when this was written. But it suffered from lack of typescript context (knowing when there are both named and default exports and omitting the re-export when there isn't one).

A new tool, [`ctix`](https://github.com/imjuni/ctix) was created that solves this problem.

This PR switches the generation of the barrel files to use ctix.

# How

Remove barrelsby, add ctix, add scripts to generate.

Previously running barrelsby required manual curation of the generated files.

Now, one can just run `yarn ctix` and the final correct files will be generated.

Note that ctix (correctly) warns when the named export and default export have the same name. This required updating StubDatabaseAdapter to just not have a default shared exported instance. Which is more correct anyways.

# Test Plan

`yarn ctx`, inspect this PR to ensure resulting barrel is the same/better (with new files from the PRs in this stack).


quinlanj (approved):

wschurman:
Grep for `barrelsby` and switch other packages over before commit.

wschurman:
### Merge activity

* **Feb 23, 10:20 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/expo/entity/255).

Issue #201, opened 2022-11-15
feature request: loadByCompositeKeyAsync


quinlanj:
https://github.com/expo/universe/pull/10823/files#r1020707597

wschurman:
Previously discussed in https://github.com/expo/entity/pull/123#issuecomment-847502513

wschurman:
> > In a subset of cases where we have unique composite keys (`UNIQUE (a, b)`), we could tell the Entity framework that the pair `(a, b)` is cacheable and have the cache invalidated when the same way single-field keys get their cache entries invalidated. I think this autoinvalidation could be built on top of the more flexible, generic secondary cacher in this PR.
> > (Autoinvalidation would not work for the case like "ORDER BY created_at DESC LIMIT 1" where there is no unique key.)
> 
> To accomplish this I think a generalized approach would be better than separate implementation. I could see it being done by factoring out what already exists in the loader into things like `loadByFieldsCompositeKey(Tuple<FieldName, FieldValue>[])` and then make `loadByFieldEqualing` the n=1 case that calls into it (and doing similar conversions for other existing methods).
> 
> The tough thing will be figuring out what the set of keys to auto-invalidate are:
> 
> * Columns: `a`, `b`, `c`
> * User specifies `a, b` as composite key
> * User updates column `b` to new value, we need to invalidate the cartesian product of old and new sub-column values (though I haven't verified this yet)

Thinking about this, let's say we have an entity definition where an entity is unique name per account:
```
TestEntity
  columns:
    id: cache - true
    other_unique_field: cache - true
    name
    account_id
  composite_keys:
    [account_id, name]
```

Before composite keys:
- Create { id: 1, other_unique_field: 'blah', name: 'hello', account_id: 1 }
   Invalidations:
     - testentity:id:1 (dataloader, cache)
     - testentity:other_unique_field:blah (dataloader, cache)
     - testentity:name:hello (dataloader)
     - testentity:account_id:1 (dataloader)
- Update { id: 1, other_unique_field: 'blah2', name: 'hello2', account_id: 1 }
   Invalidations:
     - testentity:id:1 (dataloader, cache)
     - testentity:other_unique_field:blah (dataloader, cache)
     - testentity:other_unique_field:blah2 (dataloader, cache)
     - testentity:name:hello (dataloader)
     - testentity:name:hello2 (dataloader)
     - testentity:account_id:1 (dataloader)
- Delete
   Invalidations:
     - testentity:id:1 (dataloader, cache)
[...]

PR #269, opened 2025-03-22
feat: add some utility classes and methods in preparation for composite keys

wschurman:
# Why

This adds the utility classes needed for composite field loading as described in #201.

# How

This adds a few concepts:
- `areSetsEqual` - simple utility for checking set equality
- `SerializableKeyMap` - a map where keys must be of a custom interface for checking equality.

# Test Plan

Add new tests, run them. Full coverage.

### DIFF
--- a/packages/entity/src/utils/collections/SerializableKeyMap.ts
+++ b/packages/entity/src/utils/collections/SerializableKeyMap.ts
@@ -1,0 +1,84 @@
+/**
+ * A serializable object. The serialized value is used to determine equality.
+ */
+export interface ISerializable<TSerialized> {
+  serialize(): TSerialized;
+}
+
+/**
+ * A map with keys that are serializable. Serialized keys are considered equal if their serialized values are equal.
+ * Otherwise, the map behaves like a regular Map.
+ */
+export abstract class SerializableKeyMap<TSerialized, K extends ISerializable<TSerialized>, V>
+  implements ReadonlyMap<K, V>, Map<K, V>
+{
+  protected readonly underlyingMap: Map<TSerialized, V>;
+
+  constructor(iterable?: Iterable<readonly [K, V]> | null) {
+    this.underlyingMap = new Map(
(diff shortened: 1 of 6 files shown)


wschurman:
* **#269** <a href="https://app.graphite.dev/github/pr/expo/entity/269?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/expo/entity/269?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sjchmiela (approved):

ide (approved):

packages/entity/src/utils/collections/SerializableKeyMap.ts
```diff
@@ -0,0 +1,84 @@
+/**
+ * A serializable object. The serialized value is used to determine equality.
+ */
+export interface ISerializable<TSerialized> {
+  serialize(): TSerialized;
+}
```
ide, line 6: One thought is this could be:
```suggestion
/**
 * A serializable object. The serialized value is used to determine equality.
 */
export interface ISerializable<TSerialized extends string> {
  toJSON(): TSerialized;
}
```
And then the serializer would call JSON.stringify() and deserializer would call JSON.parse(). However, sometimes we may not want JSON due to its bulkiness compared to potentially more compact formats, in which case the more generic interface makes sense,

wschurman: Precisely, for single field load values it's fine for it to just be the field name and field value. For composite, JSON is likely the most compact. This is only used in-memory (not used as part of any redis keys, for example) so it's changeable in the future too if we need to.

wschurman:
### Merge activity

* **Mar 28, 2:32 PM EDT**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/expo/entity/269).

PR #271, opened 2025-03-22
feat: convert batched/cached loader interface to holder pattern

wschurman:
# Why

Step 2 of #201.

This PR abstracts out (genericizes) the load key and load value interfaces for batched/cached loading in preparation for composite field loading.

# How

Accomplishing this requires a number of things:
1. A set of interfaces for the keys and load values to be used to load entity field objects.
2. A concrete implementation of that interface for the current use case (single fieldName, fieldValue(s)).
3. Data structure changes in the adapters and data manager to handle the new key type not being comparable via `===`. (see base PR for new map data structure for using these objects as keys).
4. Updates to the test matchers to know how to compare the serializable holder objects.

# Test Plan

Update a number of tests and add a few new tests. This has full coverage.

PR #272, opened 2025-03-22
feat: add composite field loading and caching

### LINKED ISSUE #201: feature request: loadByCompositeKeyAsync

wschurman:
# Why

This is part 3 of #201. This PR adds support for composite field loading.

Closes #201.

# How

1. Add a new composite field and composite field value holder. (holder concept introduced in parent PR)
2. Add new loader methods.
3. Fix the knex database adapter since loading via `ANY` is not possible for composite columns.

# Knex change

This turned out to be a change needed at the 11th hour after much testing, and is somewhat a pain. Previous comments on architecture noted:
> WHERE IN makes a mess of the query stats table (for each cardinality of arrays there's a separate entry). We use WHERE = ANY so that the query shape is the same whether there are 2, 3, or 100 values to match.

Unfortunately, postgres doesn't support `ANY` on these tuple like types (composite) without explicit type casting. The precise error thrown is `input of anonymous composite types is not implemented`.

I looked into using a `unnest` trick as described https://blog.jakubholy.net/2023/pg-select-where-match-on-list-of-tuples/ but it also got very messy.

So I ended up just using `WHERE IN` for the composite loads since it's the simplest.

# Test Plan

Add a lot of new tests.


wschurman:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/expo/entity/272?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#272** <a href="https://app.graphite.dev/github/pr/expo/entity/272?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/expo/entity/272?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#271** <a href="https://app.graphite.dev/github/pr/expo/entity/271?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#269** <a href="https://app.graphite.dev/github/pr/expo/entity/269?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

wschurman:
One change before commit: check for duplicate fields in composite key, also check that the size of the composite key is >= 2 fields.

packages/entity-database-adapter-knex/src/PostgresEntityDatabaseAdapter.ts
```diff
@@ -42,18 +42,45 @@ export default class PostgresEntityDatabaseAdapter<
   }
 
   protected async fetchManyWhereInternalAsync(
-    queryInterface: any,
+    queryInterface: Knex,
     tableName: string,
     tableColumns: readonly string[],
-    tableValueValues: (readonly any[])[],
```
ide, line 48: I think tableTuples or tableValuesList or tableValuesArray is clearer and more accurate.

It previously was a list where each item was a table value: tableValues.
Now it's a list where each item is table values: "tableValuess" so to speak. But I don't think it becomes ValueValues.

packages/entity-database-adapter-knex/src/PostgresEntityDatabaseAdapter.ts
```diff
@@ -42,18 +42,45 @@ export default class PostgresEntityDatabaseAdapter<
...
+      );
+    }
+
     return await wrapNativePostgresCallAsync(() =>
-      // queryInterface.select().from(tableName).whereIn(tableFields, tableFieldsValues),
-      queryInterface
-        .select()
-        .from(tableName)
-        .whereRaw(`(${rawPlaceholders}) = ANY(?)`, [...tableColumns, tableValueValues]),
+      queryInterface.select().from(tableName).whereIn(tableColumns, tableValueValues),
```
ide, line 83: Not going to block the PR on this but I'd like to compare the query plan between WHERE IN and UNNEST and if they are comparable for performance, generate the UNNEST query. We do UNNEST in other parts of the code already.

packages/entity/src/internal/CompositeFieldHolder.ts
```diff
@@ -0,0 +1,208 @@
...
+  ): CompositeFieldValueHolder<TFields, TCompositeField> {
+    return new CompositeFieldValueHolder(
+      JSON.parse(serialized) as EntityCompositeFieldValue<TFields, TCompositeField>,
+    );
+  }
+
+  public serialize(): SerializedCompositeFieldValueHolder {
+    return JSON.stringify(
+      this.compositeFieldValue,
+      Object.keys(this.compositeFieldValue).sort(),
```
ide, line 189: This call doesn't look quite right.

packages/entity-database-adapter-knex/src/PostgresEntityDatabaseAdapter.ts
wschurman: The issue with unnest is similar to the issue with `ANY` when using tuples: one must explicitly cast the tuple. For normal queries this is totally fine. But in this function (and in the entity library in general) we don't have the database-specific column types that we'd need to cast to at runtime.

Here's what a theoretical `UNNEST` implementation would look like here:

```typescript
    const tableColumnTuplePlaceholders = tableColumnTuple.map(() => '??').join(', ');
    const tableValueTuplesGroupedByColumn = new Array(tableColumnTuple.length)
      .fill(null)
      .map((_, i) => tableValueTuples.map((tableValueTuple) => tableValueTuple[i]));
    const tableValueTuplesPlaceholders = tableColumnTuple.map(() => '?::boolean[]').join(', ');
    return await wrapNativePostgresCallAsync(() =>
      queryInterface
        .select()
        .from(tableName)
        .whereRaw(
          `(${tableColumnTuplePlaceholders}) IN (SELECT * FROM UNNEST(${tableValueTuplesPlaceholders}))`,
          [...tableColumnTuple, ...tableValueTuplesGroupedByColumn],
        ),
    );
```

Notice in `tableValueTuplesPlaceholders` we explicitly have to add a cast to an array with items of a specific type. Without this cast, postgres throws an error:
> error: select * from "postgres_test_entities" where ("has_a_cat", "has_a_dog") IN (SELECT * FROM UNNEST($1, $2)) - function pg_catalog.unnest(unknown) is not unique

With an incorrect cast, the error is something like
> error: select * from "postgres_test_entities" where ("has_a_cat", "has_a_dog") IN (SELECT * FROM UNNEST($1::varchar[], $2::varchar[])) - operator does not exist: boolean = character varying

So this tells me that to correctly unnest anonymous tuples one must explicitly cast the tuple, similar to `ANY`.

packages/entity/src/internal/CompositeFieldHolder.ts
ide: I think this is the first time I've seen the replacer used that way instead of as a function. It's a good use case, thanks for adding the comment.

packages/entity-example/src/adapters/InMemoryDatabaseAdapter.ts
```diff
@@ -34,13 +34,13 @@ class InMemoryDatabaseAdapter<T extends Record<string, any>> extends EntityDatab
   protected async fetchManyWhereInternalAsync(
     _queryInterface: any,
     _tableName: string,
-    tableColumns: readonly string[],
-    tableValueValues: (readonly any[])[],
+    tableColumnTuple: readonly string[],
```
ide, line 37: This should still be tableColumns. "Tuple" to me here means the same as "row". Come to think of it, the other parameter could just be named `tableRows` or `tableTuples`.

- Column: the name of a single column
- Columns: an array, where each item is a column name
- Row or tuple: an array of values, where each item is this row's value for a specific column
- Rows or tuples: an array of rows/tuples, hence a 2D array

"Row" to me has more of a tabular connotation than "tuple", which I think is appropriate here: we have parameters that mention "table" and "column". 

In contrast, if we were thinking more abstractly and separated from database tables, I'd say "fieldNames" and "fieldValues/fieldTuples/fieldValueTuples" would be more appropriate.

wschurman: `fieldNames`/`fieldValues` refer to the entity fields already (since field names can differ from column names, often in casing). `fetchManyWhere` uses `key`/`values` which were `fieldName`/`fieldValue` in pre-composite times, and then translates them to database column names and calls this abstract method `fetchManyWhereInternalAsync` with columns. I'll go with `tableTuples`.

ide (approved):

PR #273, opened 2025-03-29
[rfc][breaking] Make ID fields cacheable by default

ide:
# Why

We've seen people repeatedly forget to add `cache: true` to ID fields, and our 10th query ranked by load was a simple "by ID" query that could have been cached. It wasn't causing database problems but would have been better to use Redis.

# How

I opted to make ID fields cacheable by default, with the option to explicitly specify `cache: false` if you want to opt out. This is a breaking change, as all implicitly uncacheable ID fields are now implicitly cacheable.

As an alternative, I considered requiring ID fields to explicitly set the `cache` option, rather than defaulting to true or false. It's easy to implement a runtime check for this.

However, to get static warnings, this requires changing the TypeScript generic type parameter of EntityConfiguration to look like `EntityConfiguration<'id', ExampleFields>`, which was too redundant for my taste. Namely, the `'id'` part is already specified in the EntityConfiguration's `idField`, and we'd be specifying `'id'` for just about every entity.

# Test Plan

Added unit tests that specific check that ID fields are automatically cacheable, and that you can opt out with `cache: false`.

### DIFF
--- a/packages/entity/src/EntityConfiguration.ts
+++ b/packages/entity/src/EntityConfiguration.ts
@@ -1,6 +1,6 @@
 import { IEntityClass } from './Entity';
 import { DatabaseAdapterFlavor, CacheAdapterFlavor } from './EntityCompanionProvider';
-import { EntityFieldDefinition } from './EntityFieldDefinition';
+import { EntityFieldDefinition, EntityFieldDefinitionOptions } from './EntityFieldDefinition';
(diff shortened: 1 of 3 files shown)


ide:
Closing this as we've opted to require the cache field to be explicit. There are cases where it's important to know that caching is enabled -- such as changing an entity schema -- so we'd want to make sure `cache: true` is known to the programmer.

PR #275, opened 2025-03-30
feat: invalidate n+/-1 cacheKeyVersion for entities for push safety

wschurman:
# Why

Please describe the motivation for this PR, and link to relevant GitHub issues, forums posts, or feature requests.

# How

How did you build this feature or fix this bug?

# Test Plan

Please describe how you tested this change and how a reviewer could reproduce your test, especially if this PR does not include automated tests!


wschurman:
* **#275** <a href="https://app.graphite.dev/github/pr/expo/entity/275?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/expo/entity/275?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

quinlanj (approved):
> While this may seem like a 3x load increase, invalidations/mutations account for a very very small percentage of entity operations in normal applications. Also this doesn't add any additional del redis calls since the call accepts a variable number of keys.

This was my main concern, but I'm glad you clarified this in the PR description!

ide (approved):
This makes sense and the implementation is clean but I am a bit hesitant about the permanent 3x increase in keys during writes and deletions. Specifically, we've seen key size affect performance and sending large numbers of key bytes has caused Redis to fail in the past for notifications.

I'd like to help get this commit through and am approving it, but I'm also wondering if there's a way to use timestamps or information about the deployment to limit how long we send 3x the keys. 99%+ of the time we aren't changing cache key versions.

wschurman:
I thought about whether there was a way to convey time/deployment-time/expiration-window and couldn't come up with anything that works in a generalized manner, or even a use-case-specific-manner for that matter. 

This PR started as a configurable entity config field called `otherCacheKeyVersionsToInvalidate`, but that only really solved the "invalidate past versions from new code" half of the problem. To solve the other half of the problem, "invalidate future versions from old code", we'd need to require deploying a preemptive change to `otherCacheKeyVersionsToInvalidate` and have that be in production for some amount of time before bumping the cache key version manually. This process just seemed too tedious to me and likely developers would just choose to risk cache poisoning in exchange for less manual overhead.

So the natural progression was trying to figure out a way do it automatically and not require too much developer overhead, which meant we always will need to invalidate at least the next version no matter what since we couldn't predict when the next version will go out. Going from that to just invalidating (x-1, x, x+1) made sense for the developer.

Here's the data for the Expo backend Redis calls:
```
mget - 16k/s (entity loads)
multi - 300/s (entity read through stores, commands below are executed in multi)
	set - 300/s 
	exec - 300/s
del - 40/s (invalidation due to mutations)
```

So, my takeaway was that even if we did 3x the del calls (though in this case we're doing the same number of del calls just with 3 keys each instead of 1), it's still be a very minuscule percentage of redis calls (0.2% -> 0.7%). But, regarding what you mention about key bytes causing issues: my sense is that since mget already is called with far more (arbitrary amount of) keys (can theoretically be called with N keys), my hope is that it will be okay to go from 1 -> 3 keys per del call since it's still a constant number of keys.

wschurman:
Maybe there's a way to automatically feed deployment age into the framework initializer (or cache adapter config)? And then use that combined with `ttlSecondsPositive`/`ttlSecondsNegative` to calculate how long to invalidate the old key for. But then we'd still have the half of the problem with the next key.

ide:
That's good to quantify the impact (0.2% -> 0.7%). It's going to be fine. It just feels wasteful to have the 3x extra traffic when it's not needed most of the time.

> we'd still have the half of the problem with the next key.

I think the old deployment would need to know when a new deployment is rolling out.

wschurman:
> I think the old deployment would need to know when a new deployment is rolling out.

That's a super interesting thought. One could conceivably make part of the deployment process for a new deployment be setting an environment variable on the old deployment telling it that it is being replaced (though technically setting an environment variable itself creates triggers a redeployment/restart which would potentially have concerns of its own during a deployment).

So the logic would be:
1. Have a boolean configuration property on the entity companion provider that instructs the framework to forward-invalidate (invalidate the next version for all entities). Set this to the value of an environment variable, `IS_DEPLOYMENT_BEING_REPLACED`.
2. Have a required date configuration property on the entity companion provider containing approximate deployment time. Set this to the value of the boot time of the pod or something like that depending on k8s/etc. Build functionality into the framework to backward-invalidate until `now() - deployment_time > max(ttlSecondsPositive, ttlSecondsNegative)`. Could instead just use a constant value to instruct the framework how long to invalidate for rather than ` max(ttlSecondsPositive, ttlSecondsNegative)`, but this would be application-specific.

I think the tradeoff is 3x extra bytes for `del` calls in exchange for less complexity in framework configuration and container-orchestration-method-agnosticism. I think I still fall on the side of choosing 3x `del` calls since it makes the default safer (pit of success). I worry that having the invalidation mechanism and configuration being complex will result in people just not implementing it since the path of least resistance is just not to do it. That being said, maybe the right design here is for this to be optional. So far Expo server implementation has fared just fine without this push safety invalidation mechanism.

wschurman:
Updated this to be configurable (explicit configuration required though since it's easy to miss). Options are:
- previous behavior (just invalidate the current version)
- surrounding versions

This allows for us to add a more complex invalidation strategy using deployment information in the future.

wschurman:
### Merge activity

* **Apr 2, 1:31 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/expo/entity/275).

PR #276, opened 2025-03-31
feat: enforce explicit id field cache property

wschurman:
# Why

Discussed #273 offline. What we settled on was that the ID field should need to explicitly specify the cache field. This way it's still possible to not cache, especially during development of a new entity at which time having a cache can be frustrating.

# How

The way this is accomplished is by using the newly-genericized id field to conditionally switching the type of field definition needed for the id field in the schema.

# Test Plan

Run all tests. See that the cache field is required by tsc in the existing entities in these packages.

### DIFF
--- a/packages/entity-database-adapter-knex/src/EntityFields.ts
+++ b/packages/entity-database-adapter-knex/src/EntityFields.ts
@@ -3,7 +3,10 @@ import { EntityFieldDefinition } from '@expo/entity';
 /**
  * EntityFieldDefinition for a Postres column with a JS JSON array type.
  */
-export class JSONArrayField extends EntityFieldDefinition<any[]> {
+export class JSONArrayField<TRequireExplicitCache extends boolean> extends EntityFieldDefinition<
+  any[],
+  TRequireExplicitCache
+> {
   protected validateInputValueInternal(value: any[]): boolean {
     return Array.isArray(value);
   }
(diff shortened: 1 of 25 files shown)


wschurman:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/expo/entity/276?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#276** <a href="https://app.graphite.dev/github/pr/expo/entity/276?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/expo/entity/276?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#275** <a href="https://app.graphite.dev/github/pr/expo/entity/275?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

ide:
When I was prototyping different approaches, I think I had a solution like this working that required the new generic type parameter that specifies the ID field (`new EntityConfiguration<ExampleFields, 'id'>(...)`) but didn't need to introduce ID-specific runtime types like `UUIDIDField()`.

wschurman:
Yeah, I'm convinced it is possible for sure. I tried for a bit, but I couldn't get it working with a test entity. I'll try a little longer to see what I can do.

ide:
I put up #279, which is similar to this PR and builds on the same base PR but doesn't define `EntityFieldDefinitionOptionsExplicitCache` for instance.

The PR is incomplete and doesn't modify packages/entity-database-adapter-knex/src/EntityFields.ts (granted, those types are probably not used as IDs; arguably BooleanField, DateField, and others are never used as IDs and could be simplified). But I think it's simpler than this PR and worth comparing.

wschurman:
### Merge activity

* **Apr 2, 1:31 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/expo/entity/276).

PR #280, opened 2025-04-07
feat: add @expo/entity-testing-utils package

wschurman:
# Why

When testing the last batch of changes in the Expo server, I noticed that I needed to use the TSMockitoExtensions. Figured it was a good idea to expose them for others to use as well.

# How

The challenge is that these should only be in dev dependencies so exposing them in the @expo/entity package is a no-go.

So, create a new package. Then, it made sense to move all the testing utilities to that package. But this requires duplicating code since it would create a dependency cycle if entity used the common code in addition to the common code using entity. So, the solution is to make the test utils depend on entity but not vice-versa and just duplicate the modules. Then, add a test to ensure the duplicated modules stay perfectly in-sync.

This also moves the other test utilities into this new package.

# Test Plan

Run all tests (full coverage).


wschurman:
* **#280** <a href="https://app.graphite.dev/github/pr/expo/entity/280?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/expo/entity/280?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

wschurman:
### Merge activity

* **Apr 9, 12:11 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/expo/entity/280).

PR #284, opened 2025-05-13
feat: add transaction-scoped dataloaders

wschurman:
# Why

This is a retry of #98. It has become more important in Expo's server application due to more things running within transactions. When it was written previously, relatively few things were transactional.

The idea is that within a transaction, subsequent reads should be able to be cached locally (dataloader) as long as the local cache is kept consistent (not in the cache adapter though).

# How

The way this is done is to keep a dataloader map (dataloader per-entity like we do for the top-level loads) per-transaction as well. Then, when in a transaction, instead of going directly to the database, go through the transaction's dataloader before hitting the database. This way loads within the transaction are coalesced and cached locally.

Invalidation becomes more complex though. 
- We still only want to invalidate the non-transactional dataloader/cache at the end of the transaction (as late as possible for least chance of race condition) for all changed entities. This didn't change.
- As for when to invalidate the transactional dataloaders, we need to think about this like database sub-transaction visibility, but only in the sense that we don't want inconsistent reads. Put another way, any load in or out of any transaction or nested transaction needs to be consistent with what is in the database, so any time the loaded value would change if it went directly do the database we must also invalidate the cache.
   - This is further complicated by the fact that postgres doesn't have real subtransactions (see #194 for more discussion) and subtransactions weren't an entity feature for the first attempt at this. This means that it's hard to test the case where a nested transaction is isolated from the parent transaction for uncommitted reads or the inverse.
   - The way I came up with achieving this is to just to over-invalidate: clear all the dataloaders in the transaction recursively starting from the outermost transaction. And do this clearing right after the mutation and at the end of the sub-transaction and at the end of the transaction (recursively).

# Test Plan

Run new tests.

### DIFF
--- a/packages/entity/src/AuthorizationResultBasedEntityMutator.ts
+++ b/packages/entity/src/AuthorizationResultBasedEntityMutator.ts
@@ -27,7 +27,93 @@ import { timeAndLogMutationEventAsync } from './metrics/EntityMetricsUtils';
 import IEntityMetricsAdapter, { EntityMetricsMutationType } from './metrics/IEntityMetricsAdapter';
 import { mapMapAsync } from './utils/collections/maps';
 
-abstract class AuthorizationResultBasedBaseMutator<
+/**
+ * Base class for entity mutators. Mutators are builder-like class instances that are
+ * responsible for creating, updating, and deleting entities, and for calling out to
+ * the loader at appropriate times to invalidate the cache(s). The loader is responsible
+ * for deciding which cache entries to invalidate for the entity being mutated.
+ *
+ * ## Notes on invalidation
+ *
+ * The primary goal of invalidation is to ensure that at any point in time, a load
+ * for an entity through a cache or layers of caches will return the most up-to-date
+ * value for that entity according to the source of truth stored in the database, and thus
+ * the read-through cache must be kept consistent (only current values are stored, others are invalidated).
+ *
+ * This is done by invalidating the cache for the entity being mutated at the end of the transaction
+ * in which the mutation is performed. This ensures that the cache is invalidated as close as possible
+ * to when the source-of-truth is updated in the database, as to reduce the likelihood of
+ * collisions with loads done at the same time on other machines.
+ *
+ * <blockquote>
+ *    The easiest way to demonstrate this reasoning is via some counter-examples.
+ *    For sake of demonstration, let's say we did invalidation immediately after the mutation instead of
+ *    at the end of the transaction.
+ *
+ *    Example 1:
+ *    - t=0. A transaction is started on machine A and within this transaction a new entity is created.
+ *           The cache for the entity is invalidated.
+ *    - t=1. Machine B tries to load the same entity outside of a transaction. It does not yet exist
+ *           so it negatively caches the entity.
+ *    - t=2. Machine A commits the transaction.
+ *    - t=3. Machine C tries to load the same entity outside of a transaction. It is negatively cached
+ *           so it returns null, even though it exists in the database.
+ *
+ *    One can see that it's strictly better to invalidate the transaction at t=2 as it would remove the
+ *    negative cache entry for the entity, thus leaving the cache consistent with the database.
+ *
+ *    Example 2:
+ *    - t=0. Entity A is created and read into the cache (everthing is consistent at this point in time).
+ *    - t=1. Machine A starts a transaction, reads entity A, updates it, and invalidates the cache.
+ *    - t=2. Machine B reads entity A outside of a transaction. Since the transaction from the step above
+ *           has not yet been committed, the changes within that transaction are not yet visible. It stores
+ *           the entity in the cache.
+ *    - t=3. Machine A commits the transaction.
+ *    - t=4. Machine C reads entity A outside of a transaction. It returns the entity from the cache which
+ *           is now inconsistent with the database.
+ *
+ *    Again, one can see that it's strictly better to invalidate the transaction at t=3 as it would remove the
+ *    stale cache entry for the entity, thus leaving the cache consistent with the database.
+ *
+ *    For deletions, one can imagine a similar series of events occurring.
+ * </blockquote>
+ *
+ * #### Invalidation as it pertains to transactions and nested transactions
+ *
+ * Invalidation becomes slightly more complex when nested transactions are considered. The general
+ * guiding principle here is that over-invalidation is strictly better than under-invalidation
+ * as far as consistency goes. This is because the database is the source of truth.
+ *
+ * For the visible-to-the-outside-world caches (cache adapters), the invalidations are done at the
+ * end of the outermost transaction (as discussed above), plus at the end of each nested transaction.
+ * While only the outermost transaction is strictly necessary for these cache adapter invalidations,
+ * the mental model of doing it at the end of each transaction, nested or otherwise, is easier to reason about.
+ *
+ * For the dataloader caches (per-transaction local caches), the invalidation is done multiple times
+ * (over-invalidation) to better ensure that the caches are always consistent with the database as read within
+ * the transaction or nested transaction.
+ * 1. Immediately after the mutation is performed (but before the transaction or nested transaction is committed).
+ * 2. At the end of the transaction (or nested transaction) itself.
+ * 3. At the end of the outermost transaction (if this is a nested transaction) and all of that transactions's nested transactions recursively.
+ *
+ * This over-invalidation is done because transaction isolation semantics are not consistent across all
+ * databases (some databases don't even have true nested transactions at all), meaning that whether
+ * a change made in a nested transaction is visible to the parent transaction(s) is not necessarily known.
+ * This means that the only way to ensure that the dataloader caches are consistent
+ * with the database is to invalidate them often, thus delegating consistency to the database. Invalidation
+ * of local caches is synchronous and immediate, so the performance impact of over-invalidation is negligible.
+ *
+ * #### Invalidation pitfalls
+ *
+ * One may have noticed that the above invalidation strategy still isn't perfect. Cache invalidation is hard.
+ * There still exists a very short moment in time between when invalidation occurs and when the transaction is committed,
+ * so dirty cache writes are still possible, especially in systems reading an object frequently and writing to the same object.
+ * For now, the entity framework does not attempt to provide a further solution to this problem since it is likely
+ * solutions will be case-specific. Some fun reads on the topic:
+ * - https://engineering.fb.com/2013/06/25/core-infra/tao-the-power-of-the-graph/
+ * - https://hazelcast.com/blog/a-hitchhikers-guide-to-caching-patterns/
+ */
+export abstract class AuthorizationResultBasedBaseMutator<
   TFields extends Record<string, any>,
   TIDField extends keyof NonNullable<Pick<TFields, TSelectedFields>>,
   TViewerContext extends ViewerContext,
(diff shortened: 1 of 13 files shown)


packages/entity/src/AuthorizationResultBasedEntityMutator.ts
```diff
@@ -282,9 +283,12 @@ export class AuthorizationResultBasedCreateMutator<
 
     const insertResult = await this.databaseAdapter.insertAsync(queryContext, this.fieldsForEntity);
 
-    queryContext.appendPostCommitInvalidationCallback(
-      entityLoader.utils.invalidateFieldsAsync.bind(entityLoader, insertResult),
-    );
+    queryContext.appendPostCommitInvalidationCallback(async () => {
+      entityLoader.utils.invalidateFieldsForTransaction(queryContext, insertResult);
+      await entityLoader.utils.invalidateFieldsAsync(insertResult);
```
quinlanj, line 288: if insertResult is already being invalidated for all txns in the tree, what is the purpose of `invalidateFieldsAsync`?

wschurman: I'll add a comment

packages/entity/src/AuthorizationResultBasedEntityMutator.ts
```diff
@@ -499,15 +504,23 @@ export class AuthorizationResultBasedUpdateMutator<
...
+    queryContext.appendPostCommitInvalidationCallback(async () => {
+      entityLoader.utils.invalidateFieldsForTransaction(
+        queryContext,
         this.originalEntity.getAllDatabaseFields(),
-      ),
-    );
-    queryContext.appendPostCommitInvalidationCallback(
-      entityLoader.utils.invalidateFieldsAsync.bind(entityLoader, this.fieldsForEntity),
+      );
+      entityLoader.utils.invalidateFieldsForTransaction(queryContext, this.fieldsForEntity);
```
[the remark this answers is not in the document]
wschurman: `fieldsForEntity` is the builder pattern record of merging the original fields with all subsequent calls to `setFields`. So it starts as the same things as the fields from the original entity and then accumulates changes such that at this point the new entity (about-to-be-committed entity) can be invalidated fully.

wschurman:
Added a long docblock about invalidation.

packages/entity/src/AuthorizationResultBasedEntityMutator.ts
```diff
@@ -27,7 +27,92 @@ import { timeAndLogMutationEventAsync } from './metrics/EntityMetricsUtils';
...
+ * the read-through cache must be kept consistent (only current values are stored, others are invalidated).
+ *
+ * This is done by invalidating the cache for the entity being mutated at the end of the transaction
+ * in which the mutation is performed. This ensures that the cache is invalidated as close as possible
+ * to when the source-of-truth is updated in the database, as to reduce the likelihood of
+ * collisions with loads done at the same time on other machines.
+ *
+ * <blockquote>
+ *    The easiest way to demonstrate this reasoning is via some counter-examples.
+ *    For sake of demonstration, let's say we did invalidation immediately after the mutation but not
```
quinlanj, line 50:
```suggestion
 *    For sake of demonstration, let's say we did invalidation immediately after the mutation but not before
```

packages/entity/src/AuthorizationResultBasedEntityMutator.ts
```diff
@@ -27,7 +27,92 @@ import { timeAndLogMutationEventAsync } from './metrics/EntityMetricsUtils';
...
+ *    Again, one can see that it's strictly better to invalidate the transaction at t=3 as it would remove the
+ *    stale cache entry for the entity, thus leaving the cache consistent with the database.
+ *
+ *    For deletions, one can imagine a similar series of events occurring.
+ * </blockquote>
+ *
+ * #### Invalidation as it pertains to transactions and nested transactions
+ *
+ * Invalidation becomes slightly more complex when nested transactions are considered. The general
+ * guiding principle here is that over-invalidation is strictly always better than under-invalidation
```
quinlanj, line 83:
```suggestion
 * guiding principle here is that over-invalidation is strictly better than under-invalidation
```

packages/entity/src/AuthorizationResultBasedEntityMutator.ts
```diff
@@ -27,7 +27,92 @@ import { timeAndLogMutationEventAsync } from './metrics/EntityMetricsUtils';
...
+ * the transaction or nested transaction.
+ * 1. Immediately after the mutation is performed (but before the transaction or nested transaction is committed).
+ * 2. At the end of the transaction (or nested transaction) itself.
+ * 3. At the end of the outermost transaction (if this is a nested transaction) and all of that transactions's nested transactions recursively.
+ *
+ * This over-invalidation is done because transaction isolation semantics are not consistent across all
+ * databases (some databases don't even have true nested transactions at all), meaning that whether
+ * a change made in a nested transaction is visible to the parent transaction(s) is not necessarily known.
+ * This means that the only way to ensure that the dataloader caches are consistent
+ * with the database is to invalidate them often, thus delegating consistency to the databse. Invalidation
```
quinlanj, line 102:
```suggestion
 * with the database is to invalidate them often, thus delegating consistency to the database. Invalidation
```

packages/entity/src/AuthorizationResultBasedEntityMutator.ts
```diff
@@ -27,7 +27,92 @@ import { timeAndLogMutationEventAsync } from './metrics/EntityMetricsUtils';
...
+ * of local caches is synchronous and immediate, so the performance impact of over-invaldation is negligible.
```
quinlanj, line 103:
```suggestion
 * of local caches is synchronous and immediate, so the performance impact of over-invalidation is negligible.
```

quinlanj (approved):
lgtm, thanks for explaining

packages/entity/src/internal/EntityDataManager.ts
```diff
@@ -236,4 +312,63 @@ export default class EntityDataManager<
...
+      );
+    } else {
+      return queryContext;
+    }
+  }
+
+  /**
+   * Get a list of all child query contexts recursively for a given query context.
+   */
+  private static getAllChildTransactionalQueryContexts(
```
ide, line 360: I would replace the word Child with Descendant because it's recursive all the way down.

packages/entity/src/internal/EntityDataManager.ts
```diff
@@ -236,4 +312,63 @@ export default class EntityDataManager<
...
+
+    // invalidate all query contexts in transaction tree
+    const outermostTransactionalQueryContext =
+      EntityDataManager.getOutermostTransactionalQueryContextIfInNestedTransaction(queryContext);
+    const allQueryContextsToInvalidate = [
+      outermostTransactionalQueryContext,
+      ...EntityDataManager.getAllChildTransactionalQueryContexts(
+        outermostTransactionalQueryContext,
+      ),
+    ];
```
ide, line 334: Just curious: Does the order matter? I think it doesn't, especially because this isn't an async function so it's not going to get interleaved with another concurrent transaction.

packages/entity/src/AuthorizationResultBasedEntityMutator.ts
```diff
@@ -27,7 +27,93 @@ import { timeAndLogMutationEventAsync } from './metrics/EntityMetricsUtils';
...
+ * responsible for creating, updating, and deleting entities, and for calling out to
+ * the loader at appropriate times to invalidate the cache(s). The loader is responsible
+ * for deciding which cache entries to invalidate for the entity being mutated.
+ *
+ * ## Notes on invalidation
+ *
+ * The primary goal of invalidation is to ensure that at any point in time, a load
+ * for an entity through a cache or layers of caches will return the most up-to-date
+ * value for that entity according to the source of truth stored in the database, and thus
+ * the read-through cache must be kept consistent (only current values are stored, others are invalidated).
```
ide, line 41: I think this doesn't hold true with this PR, but please correct me if I'm explaining something you already know (this was new to me and a misunderstanding I had):

Under Postgres's default transaction isolation level of READ COMMITTED, each statement -- not transaction! - gets its own snapshot of the database. This means:

1. Transaction 1 reads a row and sees Value 1
2. Transaction 2 writes Value 2 to the same row
3. Transaction 1 reads the row and sees Value 1 because Transaction 2 has not yet been committed.
4. Transaction 2 is committed.
5. Transaction 1 reads the row and **sees Value 2**, not Value 1. This was the part that was surprising to me.

The REPEATABLE READ and SERIALIZABLE transaction levels address this. However, they raise the likelihood of errors like "could not serialize access" if Transaction 1 were to try to write to the row.

---

With this PR, I believe Transaction 1 would still see Value 1 even after Transaction 2 is committed (and clears Redis) because of the DataLoader. I don't think that's necessarily a problem, I wanted to call it out though.

wschurman: Excellent point, and one that I hadn't thought of. The answer is yes, this would cause the behavior you describe for the default postgres transaction isolation level. Transaction 1 would still see Value 1 even after Transaction 2 is committed. And honestly this is how I thought postgres was by default. Kinda surprising to me that it isn't the default. I almost think maybe we should change our default to `repeatable read` for Expo www since I bet we're not the only people who thinks this is the case. I always assumed a "could not serialize access" was standard for conflicting read-write combinations.

https://www.postgresql.org/docs/current/transaction-iso.html is a good read for sure for everyone.

The interesting thing I have to force myself to remember though is that, theoretically at least (we don't use them for Expo www):
- the entity framework supports all transaction levels for postgres (#207)
- the entity framework theoretically could be backed by other DMBS/database adapters than postgres.

Because it's not well-specified or consistent what the behavior of things like dirty reads, nonrepeatable reads, etc across DBMSs is, and also I'd rather not fork the invalidation code based on the specified postgres transaction isolation level if we can help it (though thinking more maybe we should), I just decided to invalidate it frequently (especially since it's relatively free to do so perf-wise). But you're correct that it doesn't solve the committed-in-a-different-transaction-invalidate-in-my-transaction problem and I don't think there's a way that a local cache could have such a mechanism.

So, thinking about this more, my gut says that hopefully it isn't a problem. My guess is that in most applications, concurrent writes would usually conflict and cause the transaction to throw. But if it does become a problem, I wonder if the solution would be to change our default transaction isolation.

packages/entity/src/internal/EntityDataManager.ts
wschurman: Correct. That's how I interpret it. Probably most theoretically correct would be `self -> parent -> outermost -> descendants` but since this is running synchronously it should be fine in any order.

packages/entity/src/AuthorizationResultBasedEntityMutator.ts
ide: > my gut says that hopefully it isn't a problem

I think it won't be a problem because we use transactions to get all-or-nothing behavior with writes, rather than trying to keep a consistent view of the world with reads. And we use database constraints to enforce a lot of correctness.

There are cases where consistent reads are important, like a shopping cart application where we want to make sure to decrement the inventory stock correctly. That can't be done with a database constraint, except to make sure inventory >= 0. But I don't think we have use cases like this and if we did, the solution would be to make the entity library disable per-transaction dataloaders for these cases. These code paths should be minimal and carefully reviewed. As the PG docs say, "Applications using this level must be prepared to retry transactions due to serialization failures."

ide (approved):

packages/entity/src/AuthorizationResultBasedEntityMutator.ts
wschurman: Added configurable setting. Made test cleaner, too.

wschurman:
### Merge activity

* **May 23, 5:48 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/expo/entity/284).

PR #297, opened 2025-06-19
chore: Unify repo-wide scripts

nicknovitski:
# Why

It's simpler, and usually faster (ctix is slower since it doesn't implement parallelization).

# How

All existing commands are supported except for `yarn ctix` in packages to rebuild the barrel file only of that package.

## typescript

Use [project references](https://www.typescriptlang.org/docs/handbook/project-references.html).  A root `tsconfig.json` lists all the packages as references, each package's `tsconfig.json` extends the root `tsconfig.base.json`, and lists the workspace packages it depends on as references.

In the root, `yarn build` builds all package which have out-of-date outputs.  In a package, `yarn tsc` builds that project, and `yarn build` builds it and any packages it depends on.

Because of interactions with ctix, the packages' tsconfig files specify `["src/**/*"]` as their input, and `.` as their rootDir, meaning their output changes from `build` to `build/src`.

These commands check and build all files, including tests, so instead of a separate set of "build" tsconfig files, the test files are added to .npmignore, so they aren't included in published packages.

## eslint

Pretty straightforward, although the root `yarn lint` command now lints all matching files in the repo.

## tests

From root, `yarn test` and `integration` start one jest process which runs all tests.  Since a test will fail if its packaged dependencies aren't built, these commands run `yarn build:all --noCheck` first.  With project references, this only builds out of date projects, and extremely fast when none of them are out of date.

Individual package `test` and `integration` commands are equivalent to running `yarn <command> <path to a package>`.

## ctix

Changing the root .ctirc file to reference each package, and changing the preface script to read the packages' names from their package.json. 

# Test Plan

I ran all these commands a bunch.

The coverage decreasing is from fixing an error in the way we were tracking coverage previously, note this part in the summary:
```
  Files                  77       89      +12  
```
So after this commit, codecov included 12 more files in its coverage.  I'm guessing these are files that are not loaded in the tests of their own package, but are in other packages.


nicknovitski:
* **#297** <a href="https://app.graphite.dev/github/pr/expo/entity/297?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#298](https://github.com/expo/entity/pull/298) <a href="https://app.graphite.dev/github/pr/expo/entity/298?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#301](https://github.com/expo/entity/pull/301) <a href="https://app.graphite.dev/github/pr/expo/entity/301?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.dev/github/pr/expo/entity/297?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#299** <a href="https://app.graphite.dev/github/pr/expo/entity/299?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#300](https://github.com/expo/entity/pull/300) <a href="https://app.graphite.dev/github/pr/expo/entity/300?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

wschurman:
Re: codecov. Based on a spot check, it looks like something is wrong with the new setup though it's hard to tell what. For example, from the report (https://app.codecov.io/gh/expo/entity/pull/297/indirect-changes) it says that L455 of `packages/entity/src/EntityPrivacyPolicy.ts` is uncovered, but there is a test that tests for that exact string error message so it must be covered? "Invalid RuleEvaluationResult returned from rule"

nicknovitski:
> `yarn integration` and `yarn typedoc` seem to fail for me locally. I think once typedoc succeeds, `yarn lint` also fails since those files aren't excluded, but this one is hard to verify since I can't run typedoc.

Ah, I deleted the empty `entity-codemode/src/index.ts`, and that makes typedoc error.  That's an easy fix.

I'll look further at the coverage differences. ~~What error do you get with `integration`?~~ e: and I saw your other comment, I'll fix that too.

nicknovitski:
Thank you for finding that specific coverage problem.  I added downstack PRs to address it.

wschurman (changes requested):
Few more issues I'm running into when testing this:
- When I run `yarn test` from a subpackage directory (`packages/entity` for example), I believe it is running all the tests from all packages? It should only run the tests from that package if possible.
- When I run `yarn test src/__tests__/EnforcingEntityLoader-test.ts` from the `packages/entity` directory it also runs all the tests. It should only run that test if possible.
- Ditto with integration tests.

wschurman:
(rest of the changes look good though!)

nicknovitski:
> Few more issues I'm running into when testing this:
> 
>     * When I run `yarn test` from a subpackage directory (`packages/entity` for example), I believe it is running all the tests from all packages? It should only run the tests from that package if possible.
> 
>     * When I run `yarn test src/__tests__/EnforcingEntityLoader-test.ts` from the `packages/entity` directory it also runs all the tests. It should only run that test if possible.
> 
>     * Ditto with integration tests.

Ah, yep.  Fixed now.

nicknovitski:
I finally fixed the coverage problem: I was not uploading all of the files which codecov uses!  The total coverage is now consistent.

wschurman (approved):

wschurman (approved):
awesome stuff

nicknovitski:
### Merge activity

* **Jun 24, 4:58 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/expo/entity/297).
* **Jun 24, 4:58 PM UTC**: @nicknovitski merged this pull request with [Graphite](https://app.graphite.dev/github/pr/expo/entity/297).

PR #306, opened 2025-09-18
feat: allow manual specification of cascadingDeleteCause in delete mutator factory method

wschurman:
# Why

In the same way that the cascadingDeleteCause can be specified in application code during a call to the load factory method (which is the advanced way of creating a loader), it should be possible to specify for the updater and deleter when constructed via their factory methods.

Note that this doesn't affect the normal `Entity.deleter(entity).deleteAsync()` method of calling. Only when called in the following form is it specifiable from application code:
```typescript
this.existingEntity
      .getViewerContext()
      .getViewerScopedEntityCompanionForClass(this.entityClass)
      .getMutatorFactory()
      .forUpdate(this.existingEntity, this.queryContext, cascadingDeleteCause);
```

Concretely, this is useful in application code for bulk/batch deletions ahead of calling the parent entity's delete. Once #220 is implemented, this theoretically wouldn't be necessary, but as it currently stands, bulk/batch deletion is implemented in application code and therefore needs access to slightly lower-level entity constructs.

# How

Change arg, propagate to mutator classes, see that it's better represented as a constructor arg now.

Add tests.

# Test Plan

Run test.

### DIFF
--- a/packages/entity/src/AuthorizationResultBasedEntityMutator.ts
+++ b/packages/entity/src/AuthorizationResultBasedEntityMutator.ts
@@ -477,6 +477,7 @@ export class AuthorizationResultBasedUpdateMutator<
     databaseAdapter: EntityDatabaseAdapter<TFields, TIDField>,
     metricsAdapter: IEntityMetricsAdapter,
     originalEntity: TEntity,
+    private readonly cascadingDeleteCause: EntityCascadingDeletionInfo | null,
   ) {
     super(
       companionProvider,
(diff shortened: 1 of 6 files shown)


wschurman:
* **#306** <a href="https://app.graphite.dev/github/pr/expo/entity/306?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/expo/entity/306?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/entity/src/AuthorizationResultBasedEntityMutator.ts
```diff
@@ -517,45 +518,33 @@ export class AuthorizationResultBasedUpdateMutator<
...
-    cascadingDeleteCause: EntityCascadingDeletionInfo,
-  ): Promise<Result<TEntity>>;
-  private async updateInTransactionAsync(
-    skipDatabaseUpdate: false,
-    cascadingDeleteCause: EntityCascadingDeletionInfo | null,
-  ): Promise<Result<TEntity>>;
-  private async updateInTransactionAsync(
-    skipDatabaseUpdate: boolean,
-    cascadingDeleteCause: EntityCascadingDeletionInfo | null,
-  ): Promise<Result<TEntity>> {
```
wschurman, line 534: going to investigate making `skipDatabaseUpdate` a constructor method as well to not lose this type precision.

wschurman: quick update: doesn't make as much sense as a constructor param. going to leave it as-is for now. loss of type precision is okay due to unit test coverage.

PR #308, opened 2025-09-30
chore: fix erroneous describe.only and add lint rule to prevent it

wschurman:
# Why

In other repositories at Expo we have this lint disallow rule to avoid accidental skips like this. This PR adds the same rule.

# How

1. Add rule.
2. Fix violation.

# Test Plan

`yarn lint`

PR #310, opened 2025-09-30
fix: validate that UUID field is lower-case

wschurman:
# Why

Technically upper-case UUIDs are valid according to https://datatracker.ietf.org/doc/html/rfc4122.

But the way entity is written expects case sensitivity due to them being treated as simple strings. For example, this would be the expected behavior:
```
const entities = await TestEntity.loader(vc).loadManyByFieldEqualingAsync('id', ['018EBFDA-DC80-782D-A891-22A0AA057D52']);
=> Map(['018EBFDA-DC80-782D-A891-22A0AA057D52', TestEntity[...]])
```

The issue is that Postgres stores them in a case-insensitive manner (which is correct according to the UUID spec). So in actuality, the above code would theoretically do something like:
```
const entities = await TestEntity.loader(vc).loadManyByFieldEqualingAsync('id', ['018EBFDA-DC80-782D-A891-22A0AA057D52']);
=> Map(['018ebfda-dc80-782d-a891-22a0aa057d52', TestEntity[...]])
```
(it actually fails due to this change of case further up in the stack at https://github.com/expo/entity/blob/077d53cd86d4c54132ea2ebcc59d1d75e2b507d0/packages/entity/src/EntityDatabaseAdapter.ts#L182, but the reason is the same: postgres transforms to lower case under-the-hood)

This results in breaking invariants the framework provides (all keys in === all keys out).

This PR solves this problem by just enforcing UUID casing in fields to match what Postgres already does to avoid this issue.

Ref: WWW-37

# How

I first tried a field normalization approach in #309, but this also had the same problem: the fields in !== fields out. Unless we were to invert the normalization on the way out, but that comes with a lot of added complexity.

So this is the solution I propose: just don't allow any upper-case characters in the UUIDs. This is technically not UUID spec compliant since we no longer accept all technically valid UUIDs.

# Test Plan

Run the test.

### DIFF
--- a/packages/entity/src/EntityFields.ts
+++ b/packages/entity/src/EntityFields.ts
@@ -6,7 +6,7 @@ import {
 
 // Use our own regex since the `uuid` package doesn't support validating UUIDv6/7/8 yet
 const UUID_REGEX =
-  /^(?:[0-9a-f]{8}-[0-9a-f]{4}-[1-8][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}|00000000-0000-0000-0000-000000000000|ffffffff-ffff-ffff-ffff-ffffffffffff)$/i;
+  /^(?:[0-9a-f]{8}-[0-9a-f]{4}-[1-8][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}|00000000-0000-0000-0000-000000000000|ffffffff-ffff-ffff-ffff-ffffffffffff)$/;
 
 /**
  * EntityFieldDefinition for a column with a JS string type.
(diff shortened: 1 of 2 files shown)


wschurman:
* **#310** <a href="https://app.graphite.dev/github/pr/expo/entity/310?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/expo/entity/310?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#308** <a href="https://app.graphite.dev/github/pr/expo/entity/308?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#309](https://github.com/expo/entity/pull/309) <a href="https://app.graphite.dev/github/pr/expo/entity/309?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

ide:
Is there a place to document it? The limitation is fine IMO as long as it's documented since it deviates from the spec.

wschurman:
### Merge activity

* **Sep 30, 7:48 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/expo/entity/310).
* **Sep 30, 7:50 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/expo/entity/310) rebased this pull request as part of a merge.
* **Sep 30, 7:52 PM UTC**: @wschurman merged this pull request with [Graphite](https://app.graphite.dev/github/pr/expo/entity/310).

PR #319, opened 2025-11-11
chore: add docblock for orderBy and ordering

wschurman:
# Why

These were missing docblocks, and are something I often have to look up.

Closes ENG-18067.

# How

Add docblocks.

# Test Plan

tsdoc

PR #318, opened 2025-11-11
chore: update other sibling dependencies

wschurman:
# Why

Part 2 of #317, just updates more dependencies. These I check the changelogs for though to be sure it isn't going to cause any issues.

# How

Read changelogs, update.

# Test Plan

CI

Changelogs:
https://github.com/redis/ioredis/blob/main/CHANGELOG.md
https://github.com/brianc/node-postgres/blob/master/CHANGELOG.md
https://github.com/beaugunderson/ip-address/compare/v10.0.0...v10.0.1

PR #320, opened 2025-11-11
feat!: separate out validators into beforeCreateAndUpdate and add new beforeDelete validators

wschurman:
# Why

For a while in Expo application code, we've used beforeDelete triggers to approximate beforeDelete validators, but we finally found a need to separate them out. Due to the nature of validators (no side-effects) vs triggers (side-effectful), there can be a need to manually run validators but not triggers. One such need is ahead of batch deletion, to ensure that an object and it's related objects can be deleted and retain any constraints implemented in validator logic.

Closes ENG-18066.

# How

Split out definition to be more similar to triggers, update tests.

# Test Plan

Add new test for these, ensure that they are called appropriately.


wschurman:
* **#320** <a href="https://app.graphite.com/github/pr/expo/entity/320?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/entity/320?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#319** <a href="https://app.graphite.com/github/pr/expo/entity/319?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#318** <a href="https://app.graphite.com/github/pr/expo/entity/318?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#317** <a href="https://app.graphite.com/github/pr/expo/entity/317?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

byronkarlen (approved):

byronkarlen:
Would there ever be a reason to separate out beforeCreate and beforeUpdate validators? Couldn't really think of a scenario but was curious if you can think of any situations where we might have wanted this.

wschurman:
It's a valid question for sure. I definitely should have explained either in the summary or docblock.

My reasoning is that validation code for objects that exist should be idempotent and consistent in both create and update situations. Put another way, validation logic should be used to keep application constraints maintained, and running it only on create or update doesn't make sense since that would mean that the application constraint could be violated. By separating them I worry it could be too easy of a foot gun.

That being said, some validators in Expo application code check `mutationInfo.type` to fork between create and update and have special validations for objects that evolve over time by design (state changes in objects). This is expected, and generally I think this logic makes sense all in one validator (even with the forks) rather than across two validators. So, that's why I think it's better to keep them one.

Deletion differs in that it is more like a precondition-checker to ensure that application constraints that the existence of the object is satisfying aren't broken by deletion of the object. So it's more of a "check the state of the whole system assuming this will be deleted" instead of the create/update "check that this object is valid internally to itself in addition to the whole system".

Practically speaking, if we were to split it into two, the migration would be a bit wonky as well:
```diff
- mutationValidators: [new Validator1(), new Validator2]
+ mutationValidators: { beforeCreate: [new Validator1(), new Validator2], beforeUpdate: [new Validator1(), new Validator2] }
```

If we do find a good reason to split, it's an easy enough migration to do at a future point as well.

wschurman:
Going to keep it as `beforeCreateAndUpdate` for now, can refactor lazily (split it) if and when we need it. I think it's hard to predict though. The fact that we just now discovered a need for on-delete validators might indicate something bigger than needs refactoring. Maybe deletion validators should be a completely separate concept?

This could also tie in to needing a more general batch-deletion and batch-update set of abstractions for entity, since the use case for deletion validators is checking validations ahead of batch (non-transactional) entity deletions.

wschurman:
Requesting re-review in case more discussion is warranted!

byronkarlen (approved):

byronkarlen:
Gotcha your comment makes sense to me.

wschurman:
### Merge activity

* **Nov 12, 4:54 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/expo/entity/320).
* **Nov 12, 5:04 PM UTC**: [Graphite](https://app.graphite.com/github/pr/expo/entity/320) rebased this pull request as part of a merge.
* **Nov 12, 5:08 PM UTC**: @wschurman merged this pull request with [Graphite](https://app.graphite.com/github/pr/expo/entity/320).

PR #322, opened 2025-11-26
feat: switch local cache to an interface

wschurman:
# Why

We've fallen too far behind on lru-cache. We previously decided to lock it @ 6.x in https://github.com/expo/entity/pull/170 because our use case couldn't pre-allocate all the data structures ahead of time efficiently.

But while reading the thread for that, I noticed the author has a ttl-cache: https://github.com/isaacs/node-lru-cache/issues/208#issuecomment-1092140956. This is much closer to our needs. 

We essentially simulate a LRU cache by updating the TTL on each get: https://github.com/isaacs/ttlcache?tab=readme-ov-file#cachegetkey-updateageonget-checkageonget-ttl--

Because our TTL is short yet we require storing N caches this use case is more optimal.

Ref: ENG-18193

# How

Switch to ttl-cache but only as a dev dependency, make exported API an interface so that consumers can choose between old lru-cache and ttl-cache.

# Test Plan

Run all tests.

### DIFF
--- a/packages/entity-cache-adapter-local-memory/package.json
+++ b/packages/entity-cache-adapter-local-memory/package.json
@@ -28,13 +28,12 @@
   "author": "Expo",
   "license": "MIT",
   "dependencies": {
-    "@expo/entity": "workspace:^",
-    "lru-cache": "^6.0.0"
+    "@expo/entity": "workspace:^"
   },
   "devDependencies": {
     "@expo/entity-testing-utils": "workspace:^",
+    "@isaacs/ttlcache": "^2.1.3",
     "@jest/globals": "^30.2.0",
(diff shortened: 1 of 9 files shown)


wschurman:
* **#323** <a href="https://app.graphite.com/github/pr/expo/entity/323?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#322** <a href="https://app.graphite.com/github/pr/expo/entity/322?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/entity/322?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #326, opened 2026-01-07
chore: provide node and yarn with mise

nicknovitski:
# Why

It's simpler, and makes it easier to run all the tests locally.

e: Also I see it's over 30% faster, from the parallelization.

# Test Plan

Everyone can run `mise run ci` to run all the same checks that CI runs, in parallel.  Also I changed the CI workflows to also run `typedoc` on PRs, to make sure that still works.


nicknovitski:
@wschurman How about now?

wschurman:
Sure, worth a try, rebase and I'll try on my machine again.

PR #369, opened 2026-01-10
fix: Inline map().filter() loops in EntityLoaderUtils

ide:
## Summary
- Replaces two instances of `.map(...).filter(x => x !== null)` with for...of loops
- Avoids allocating intermediate arrays with null values that are immediately filtered out

## Benchmarks
| Implementation | ops/sec | Speedup |
|----------------|---------|---------|
| map().filter() | 7.0M | baseline |
| for loop | 12.7M | **1.8x** |

Also reduces allocations: 8 objects per call → 7 objects per call.

## Test plan
- [x] All existing tests pass (362/362)


wschurman (approved):
Updated the PR title for this as well for conventional commits (see other one).

PR #371, opened 2026-01-10
Cache sorted keys in CompositeFieldValueHolder constructor (1.29x faster)

ide:
## Summary
Cache the sorted keys array in the constructor rather than recomputing `Object.keys().sort()` on every `serialize()` call.

The `compositeFieldValue` is `readonly` and its keys cannot change after construction, making this caching safe.

## Benchmark Results
For pre-constructed instances, serialize() is **1.29x faster**:

| Scenario | Original | Optimized | Speedup |
|----------|----------|-----------|---------|
| serialize() 100K times | 30.08ms | 23.30ms | 1.29x |
| 5 fields, serialize 5x | 291.78ms | 225.95ms | 1.29x |

## Test plan
- [x] All 470 existing tests pass
- [x] Serialization output is identical (verified in benchmark)
- [x] Key order independence preserved (different input order → same output)


ide:
Closing - the optimization doesn't provide real-world benefit since we never call serialize() more than once on the same CompositeFieldValueHolder instance.

PR #372, opened 2026-01-10
Precompute field-to-transformer map during adapter construction (1.36x faster)

ide:
## Summary
- Build a precomputed `fieldName → transformer` map once during adapter construction
- Replace two map lookups per field (schema + fieldTransformerMap) with a single direct lookup
- Maintains backward compatibility by making the precomputed map optional in transformation utility functions

## Benchmark Results
For full object transformation (15 fields), **1.36x faster**:

| Implementation | Time for 100K iterations |
|----------------|--------------------------|
| Original | 88.35ms |
| Precomputed | 64.83ms |

## Test plan
- [x] All 470 tests pass

### DIFF
--- a/packages/entity/src/EntityDatabaseAdapter.ts
+++ b/packages/entity/src/EntityDatabaseAdapter.ts
@@ -10,8 +10,10 @@ import {
   EntityDatabaseAdapterExcessiveUpdateResultError,
 } from './errors/EntityDatabaseAdapterError';
 import {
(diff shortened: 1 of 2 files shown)


ide:
Closing - the optimization adds API complexity (passing both fieldTransformerMap and precomputedFieldTransformerMap) for negligible real-world benefit. Skipping one Map.get() per field doesn't matter when database I/O dominates.

PR #376, opened 2026-01-12
feat: synthetically set cascading set-null fields to null during canViewerDeleteAsync evaluation

wschurman:
# Why

Recently in the Expo server, we found ourselves writing the conditional privacy policy rule that allows an entity to be updated if the update is a nullification of a field during a cascade deletion. 

This works during the real deletion: the field is nullified ahead of privacy policy execution: https://github.com/expo/entity/blob/main/packages/entity/src/AuthorizationResultBasedEntityMutator.ts#L1014.

The problem is that `canViewerDeleteAsync` doesn't do the same nullification ahead of privacy policy evaluation since `canViewerDeleteAsync` doesn't do any actual mutations.

Closes ENG-18502.

# How

The solution is to synthetically update the field in the entity being cascade-set-null-ed ahead of privacy policy evaluation and to provide it within the `evaluationContext` synthetically.

# Test Plan

Run new test.

### DIFF
--- a/packages/entity/src/utils/EntityPrivacyUtils.ts
+++ b/packages/entity/src/utils/EntityPrivacyUtils.ts
@@ -5,9 +5,9 @@ import {
   EntityEdgeDeletionAuthorizationInferenceBehavior,
   EntityEdgeDeletionBehavior,
 } from '../EntityFieldDefinition';
-import { EntityCascadingDeletionInfo } from '../EntityMutationInfo';
-import { EntityPrivacyPolicy } from '../EntityPrivacyPolicy';
+import { EntityPrivacyPolicy, EntityPrivacyPolicyEvaluationContext } from '../EntityPrivacyPolicy';
 import { EntityQueryContext } from '../EntityQueryContext';
+import { ReadonlyEntity } from '../ReadonlyEntity';
 import { ViewerContext } from '../ViewerContext';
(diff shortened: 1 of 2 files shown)


wschurman:
* **#376** <a href="https://app.graphite.com/github/pr/expo/entity/376?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/entity/376?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/entity/src/utils/EntityPrivacyUtils.ts
```diff
@@ -435,14 +447,43 @@ async function canViewerDeleteInternalAsync<
...
+
+              const allFields = entity.getAllDatabaseFields();
+              const syntheticFields = {
+                ...allFields,
+                [fieldName]: null,
+              };
+
+              return {
+                previousValue: entity,
+                syntheticallyUpdatedVale: entityLoader.utils.constructEntity(syntheticFields),
```
sjchmiela:
```suggestion
                syntheticallyUpdatedValue: entityLoader.utils.constructEntity(syntheticFields),
```

sjchmiela (approved):
🎉 🎉 🎉

PR #399, opened 2026-01-23
feat: Add AllowIfInParentCascadeDeletionPrivacyPolicyRule

wschurman:
# Why

Found this to be useful in the Expo backend application. The use case is that one may wish to permit access for the mostly-programmatic step of nullification of a field in an entity, whether or not any of the other privacy policy rules in the policy allow.

For example, let's say a user creates a build on organization X. Their user ID may be recorded in a foreign key field on the build so that the application can query all builds for that user. But then the user loses access to organization X. Then, they want to delete their user. This will do a cascade set null on the builds initiated by the user, but the user no longer has permission to update those builds via the normal organization membership permission rules, so this explicit allow-if-cascade-rule is used.

# How

Add a generic rule for use in applications.

# Test Plan

Add test.

### DIFF
--- a/packages/entity/src/index.ts
+++ b/packages/entity/src/index.ts
@@ -66,6 +66,7 @@ export * from './internal/SingleFieldHolder';
 export * from './metrics/EntityMetricsUtils';
 export * from './metrics/IEntityMetricsAdapter';
 export * from './metrics/NoOpEntityMetricsAdapter';
+export * from './rules/AllowIfInParentCascadeDeletionPrivacyPolicyRule';
 export * from './rules/AlwaysAllowPrivacyPolicyRule';
 export * from './rules/AlwaysDenyPrivacyPolicyRule';
 export * from './rules/AlwaysSkipPrivacyPolicyRule';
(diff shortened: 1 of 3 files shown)


wschurman:
* **#401** <a href="https://app.graphite.com/github/pr/expo/entity/401?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#400** <a href="https://app.graphite.com/github/pr/expo/entity/400?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#399** <a href="https://app.graphite.com/github/pr/expo/entity/399?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/entity/399?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

wschurman:
### Merge activity

* **Jan 27, 6:07 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/expo/entity/399).
* **Jan 27, 6:07 AM UTC**: @wschurman merged this pull request with [Graphite](https://app.graphite.com/github/pr/expo/entity/399).

PR #414, opened 2026-02-01
feat!: Add sql template and loader method

wschurman:
# Why

This is the first new feature afforded by the changes in #407.

It adds a feature that is fairly common amongst typescript ORMs: sql via tagged templates for safe value interpolation into SQL statements.

# How

Quite a lot of ORMs have this:
- https://github.com/drizzle-team/drizzle-orm
- https://github.com/mikro-orm/mikro-orm
- https://github.com/blakeembrey/sql-template-tag
- https://github.com/andywer/squid
- https://github.com/Seb-C/kiss-orm

This adds support for it to entity knex.

See tests for examples, especially `PostgresEntityIntegration-test.ts`.

The idea is that this will replace `loadManyByRawWhereClauseAsync` over the course of a few versions.

# Test Plan

Full test coverage.

PR #413, opened 2026-02-01
feat: add codemod for upcoming v0.55.0-v0.56.0

wschurman:
# Why

This adds a codemod for #407-#410 changes.

# How

Replace `.loader` calls with `.knexLoader` where applicable.

# Test Plan

Run the test.

PR #412, opened 2026-02-01
feat: add entity-database-adapter-knex-testing-utils containing StubPostgresDatabaseAdapter

wschurman:
# Why

As of moving raw and conjunction loading into the `entity-database-adapter-knex` package, the stub database adapater is no longer as useful in applications that use the package (it used to support conjunction loading synthetically for unit tests).

# How

This PR adds a new package, `entity-database-adapter-knex-testing-utils` that provides what StubDatabaseAdapter used to be (it has conjunction loading).

The idea is that applications that use `entity-database-adapter-knex` will use this package for their stub database needs and unit test companion provider.

# Test Plan

Full test coverage.

PR #410, opened 2026-02-01
feat!: move knex loaders into entity-database-adapter-knex package

wschurman:
# Why

Part 3 of #407.

# How

This is the main PR of the stack. It creates a mechanism for "installing" the extension methods on required objects, and then installs them and tests them.

The goal is to make it better self-contained about what is core in the library and guaranteed to be efficient (dataloader'd and cached) and what is not (raw SQL and SQLFragment loads in a future PR).

This is becoming more critical in a time of agentic coding where having a discrete set of loaders that can be linted separately and instruct agents to run `EXPLAIN ANALYZE` in raw cases.

# Test Plan

Full test coverage.

PR #422, opened 2026-02-05
feat: Add paginated loader to entity-database-adapter-knex

wschurman:
# Why

This is the main goal of the stack that started in #407. It adds a paginated loader for entities for applications that use the knex postgres database adapter.

# How

Many many iterations of claude.

The general principle is hoisted from Expo server application code: fetch `limit + 1` entities to know if there are more pages.

This makes use of the sql loader added in #414 to do the actual page load. Exact implementation described in docblocks.

Next PRs:
- Add `includeTotal` option and use postgres windows if possible `SELECT *, COUNT(*) OVER() as total_count`
- Add search (ilike and trigram)

# Test Plan

Full test coverage and example usage in tests.


wschurman:
* **#431** <a href="https://app.graphite.com/github/pr/expo/entity/431?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#430** <a href="https://app.graphite.com/github/pr/expo/entity/430?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#422** <a href="https://app.graphite.com/github/pr/expo/entity/422?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/entity/422?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#429** <a href="https://app.graphite.com/github/pr/expo/entity/429?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#425** <a href="https://app.graphite.com/github/pr/expo/entity/425?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#424** <a href="https://app.graphite.com/github/pr/expo/entity/424?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#414** <a href="https://app.graphite.com/github/pr/expo/entity/414?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#413** <a href="https://app.graphite.com/github/pr/expo/entity/413?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#412** <a href="https://app.graphite.com/github/pr/expo/entity/412?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#411** <a href="https://app.graphite.com/github/pr/expo/entity/411?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#410** <a href="https://app.graphite.com/github/pr/expo/entity/410?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#408** <a href="https://app.graphite.com/github/pr/expo/entity/408?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#409](https://github.com/expo/entity/pull/409) <a href="https://app.graphite.com/github/pr/expo/entity/409?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#407** <a href="https://app.graphite.com/github/pr/expo/entity/407?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/entity-database-adapter-knex/src/internal/EntityKnexDataManager.ts
```diff
@@ -106,4 +174,211 @@ export class EntityKnexDataManager<
...
+    if (!hasId) {
+      return [...clauses, { fieldName: idField, order: OrderByOrdering.ASCENDING }];
+    }
+    return clauses;
+  }
+
+  private encodeCursor(fieldObject: Readonly<TFields>, fields: readonly (keyof TFields)[]): string {
+    const cursorData: Partial<TFields> = {};
+    for (const field of fields) {
+      cursorData[field] = fieldObject[field];
```
quinlanj: This works for most fields but if we want to support the Date field it will be trickier because Postgres stores dates with a finer granularity, and it gets converted to the JS date when it comes out of the DB at a rougher granularity. This is a big problem when you rely on the Date field to order the objects.

We'll either have to encode the date as a string to preserve granularity or have an option to encode the cursor with just the entity ID and lookup the cursor in a  `WITH ....` statement appended to the main SQL so we never expose the Date in JS.

ide: > encode the cursor with just the entity ID and lookup the cursor

I believe this is what we do in some parts of the code. A cursor is a entity ID showing the last entity we fetched.

ide (commented):
Quin mentioned specifying an entity ID alone as a cursor. I think this is a pattern we will want to use often as it's compact and stable.

wschurman:
Updated to use ID alone as the opaque external cursor, and then a sub-query to derive the postgres cursor. A similar approach was already in progress for search trigram similarity so this makes them consistent.

wschurman:
(requesting re-review, easiest to see the diff on graphite by using the versions compare. no tests needed to change which is a good sign)

PR #437, opened 2026-02-13
fix: pass in augmentable classes to installation

wschurman:
# Why

An idea proposed by @ide in https://github.com/expo/entity/pull/410#pullrequestreview-3765893898.

This makes it more testable and clearer about what objects installations are "allowed" to augment.

# How

Pass in the classes explicitly.

# Test Plan

`yarn tsc`

PR #436, opened 2026-02-13
feat: Add pagination max page size configuration

wschurman:
# Why

Currently, the Entity pagination system allows clients to request arbitrarily large page sizes using the `first` or `last` parameters in pagination queries. This can lead to:
- Performance degradation when clients request excessively large pages
- Potential memory issues on the server
- Unpredictable query response times
- Difficulty in capacity planning and resource allocation

By adding configurable max page size limits, we can:
- Protect backend services from resource exhaustion
- Ensure predictable performance characteristics
- Provide better control over API usage patterns
- Maintain backward compatibility (no limit by default)

# How

The implementation adds a `paginationMaxPageSize` configuration option to the PostgresEntityDatabaseAdapterProvider that flows through the adapter hierarchy:

1. **Configuration Interface**: Added `PostgresEntityDatabaseAdapterConfiguration` with an optional `paginationMaxPageSize` field
2. **Provider Level**: `PostgresEntityDatabaseAdapterProvider` constructor now accepts the configuration object
3. **Adapter Level**:
   - `BasePostgresEntityDatabaseAdapter` defines a `paginationMaxPageSize` getter (defaults to `undefined`)
   - `PostgresEntityDatabaseAdapter` overrides it to return the configured value
4. **Validation**: `EntityKnexDataManager.loadPageAsync` validates `first`/`last` parameters against the max page size before executing queries

The validation throws descriptive errors when limits are exceeded:
- `"first must not exceed maximum page size of {limit}"`
- `"last must not exceed maximum page size of {limit}"`

Usage example:
```typescript
const provider = new PostgresEntityDatabaseAdapterProvider({
  paginationMaxPageSize: 100
});
```

# Test Plan

Added comprehensive unit tests in `EntityKnexDataManager-test.ts`.

PR #441, opened 2026-02-13
fix: refactor install method to instead be free functions

wschurman:
# Why

Asked claude extended thinking to analyze the `installExtensions` monkey-patching strategy we went with in  #410 and compare it to other typescript ORM libraries.

It said essentially that it was ok, but most modern libraries choose not to do it that way, and instead recommended "free functions" (TIL about that term) to instantiate the loaders. (`knexLoader(BlahEntity, vc)`)

Working with it even more, I came up with an even better method on top of free functions to keep the `BlahEntity.knexLoader(vc)` syntax. The strategy is to:
1. Convert to free functions (this PR).
2. Add `PostgresEntity`/`ReadonlyPostgresEntity` subclasses that application entities can extend that implement the static extension methods.
3. Revert the codemod change from this PR to keep the codemod to prefer the `BlahEntity.knexLoader(vc)` syntax since most applications should use the subclasses (though it's not strictly necessary if they prefer the free function at callsites).
4. Revert the codemod in the integration tests to express better syntax pattern.

# How

Claude did most of this, but broadly speaking the strategy is to have the free functions, and then instead of patching the prototypes/classes for singleton caching, use a WeakMap where the core package singleton is the key and the value is the new knex-specific object.

# Test Plan

Run all tests.


wschurman:
* **#444** <a href="https://app.graphite.com/github/pr/expo/entity/444?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#446](https://github.com/expo/entity/pull/446) <a href="https://app.graphite.com/github/pr/expo/entity/446?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#448](https://github.com/expo/entity/pull/448) <a href="https://app.graphite.com/github/pr/expo/entity/448?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#443** <a href="https://app.graphite.com/github/pr/expo/entity/443?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#442** <a href="https://app.graphite.com/github/pr/expo/entity/442?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#441** <a href="https://app.graphite.com/github/pr/expo/entity/441?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/entity/441?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#438** <a href="https://app.graphite.com/github/pr/expo/entity/438?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#437** <a href="https://app.graphite.com/github/pr/expo/entity/437?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#436** <a href="https://app.graphite.com/github/pr/expo/entity/436?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#431** <a href="https://app.graphite.com/github/pr/expo/entity/431?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

ide:
The main tradeoff that comes to mind is (lack of) multiple inheritance. Maybe it doesn't matter because entities are backed by only one type of database.

wschurman:
It's a valid tradeoff and one that if it is important enough for an application they can opt to just use the free functions added here instead of the base classes (they're equivalent as shown by the tests in the PR that adds the subclasses). For Expo's server, we use subclasses already for all our entities. 

It'd be cool if mixins were officially supported in TS/JS. The other option is a higher-order component function (closest thing to mixins), something like:
```
class BlahEntity extends Entity { ... }

export default withKnexLoaders(BlahEntity)
```

But the key is that these approaches all require the free functions added here, and for now the subclass approach matches most closely to what the Expo server will need.

wschurman:
### Merge activity

* **Feb 16, 8:58 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/expo/entity/441).
* **Feb 16, 9:11 PM UTC**: [Graphite](https://app.graphite.com/github/pr/expo/entity/441) rebased this pull request as part of a merge.
* **Feb 16, 9:14 PM UTC**: @wschurman merged this pull request with [Graphite](https://app.graphite.com/github/pr/expo/entity/441).

PR #452, opened 2026-02-16
fix!: change adaptor -> adapter for consistency

wschurman:
# Why

As part of #451, claude noticed that we used both american and british english variants of "adapter"/"adaptor". This makes them consistent by making them all "adapter".

# How

Update name. Technically this is a breaking change but the nature of the library allows `tsc` to catch it in all application code that uses the library.

# Test Plan

`yarn tsc`

PR #451, opened 2026-02-16
chore: docblock improvements

wschurman:
# Why

Asked claude to audit the docblocks for incorrect or missing info. This is what it mentioned as being good improvements.

# How

Update docblocks.

# Test Plan

Proofread.

PR #448, opened 2026-02-16
fix: eliminate some unnecessary any casts

wschurman:
# Why

Did a brief audit of `any` casts in the library and removed ones that are no longer needed or could be replaced with a better concrete type.

# How

Grep for `as any`, fix.

# Test Plan

`yarn tsc`

PR #453, opened 2026-02-17
fix: explicitly document behavior when cursor entity no longer exists

wschurman:
# Why

The one downside of using id-based cursor pagination and subqueries (added in #422, #431) is that it risks the row referenced by the cursor being deleted between pagination requests. It's an edge case though to be clear. Encoding the full cursor alleviates this, but has it's own downsides since trigram similarity isn't encodable and neither are things like Date objects. It is still preferable to use ID-based cursors for all pagination.

But it becomes the library's responsibility to explicitly document what the behavior is when a row referenced by the cursor is no longer present. This PR does this.

# How

When the cursor row is no longer present, the tuple evaluates to `NULL`, which produces a empty page.

The alternative to this behavior is to run an id check query ahead of the pagination query, and throw an error if it doesn't exist, or even return the first page of data if it doesn't exist. Both of these are less optimal since they cause unexpected behavior during results consumption.

So we keep the behavior as is and explicitly document it.

# Test Plan

Run new tests.

### DIFF
--- a/packages/entity-database-adapter-knex/src/__integration-tests__/PostgresEntityIntegration-test.ts
+++ b/packages/entity-database-adapter-knex/src/__integration-tests__/PostgresEntityIntegration-test.ts
@@ -1618,8 +1618,6 @@ describe('postgres entity integration', () => {
         );
 
         // Create test data with names that sort in a specific order
-        await PostgresTestEntity.dropPostgresTableAsync(knexInstance);
-        await PostgresTestEntity.createOrTruncatePostgresTableAsync(knexInstance);
 
         const entities = [];
         for (let i = 1; i <= 5; i++) {
@@ -1673,8 +1671,6 @@ describe('postgres entity integration', () => {
         );
 
         // Create entities with duplicate values to test stability
-        await PostgresTestEntity.dropPostgresTableAsync(knexInstance);
-        await PostgresTestEntity.createOrTruncatePostgresTableAsync(knexInstance);
 
         const entities = [];
         for (let i = 1; i <= 6; i++) {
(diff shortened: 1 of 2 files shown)


wschurman:
* **#453** <a href="https://app.graphite.com/github/pr/expo/entity/453?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/entity/453?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#448** <a href="https://app.graphite.com/github/pr/expo/entity/448?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#452** <a href="https://app.graphite.com/github/pr/expo/entity/452?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#451** <a href="https://app.graphite.com/github/pr/expo/entity/451?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#454** <a href="https://app.graphite.com/github/pr/expo/entity/454?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

ide (approved):

wschurman:
### Merge activity

* **Feb 17, 10:40 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/expo/entity/453).
* **Feb 17, 10:53 PM UTC**: [Graphite](https://app.graphite.com/github/pr/expo/entity/453) rebased this pull request as part of a merge.
* **Feb 17, 10:56 PM UTC**: @wschurman merged this pull request with [Graphite](https://app.graphite.com/github/pr/expo/entity/453).

PR #464, opened 2026-02-24
feat: Add authorization action to EntityPrivacyPolicyRuleEvaluationContext

wschurman:
# Why

When a rule is used across multiple authorization actions and the rule wants to have specific behavior for a certain authorization action, writing the code is made simpler by passing the action to the rule evaluation itself. In the past, this required a constructor parameter.

For example, let's say there's a rule that always allows unless the viewerContext is named "bob", but only should disallow bob from doing mutations. It's easiest to write the rule as:
```
PrivacyPolicyRule
  evaluateAsync
    if (viewerContext.name === "bob" && evaluationContext.action !== READ) {
      return SKIP
```

Replaces #462 which added it as a param. This is cleaner.

# How

Add action to the context.

# Test Plan

Run new test.

### DIFF
--- a/packages/entity-testing-utils/src/PrivacyPolicyRuleTestUtils.ts
+++ b/packages/entity-testing-utils/src/PrivacyPolicyRuleTestUtils.ts
@@ -1,10 +1,10 @@
 import {
-  EntityPrivacyPolicyEvaluationContext,
   EntityQueryContext,
   ReadonlyEntity,
   ViewerContext,
   PrivacyPolicyRule,
   RuleEvaluationResult,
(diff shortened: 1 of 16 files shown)


wschurman:
* **#464** <a href="https://app.graphite.com/github/pr/expo/entity/464?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/entity/464?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ide (approved):

PR #476, opened 2026-02-26
fix: correct precedence parentheses for SQLFragment joins

wschurman:
# Why

There's a bug in where clause construction, which used SQLFragmentHelpers.and: https://github.com/expo/entity/blob/main/packages/entity-database-adapter-knex/src/internal/EntityKnexDataManager.ts#L303

The bug is that given something like:
```
where: sql`account_id = ${accountId}`,
searchWhere: sql`something ILIKE term OR similarity(term, ...) OR ...`
```
it would produce
```
WHERE account_id = ${accountId} AND something ILIKE term OR similarity(term, ...) OR ...
```

which has the incorrect parentheses and therefore incorrect precedence.

This exposed more bugs in the precedence in `or`, `and`, `join`, etc methods.

# How

The fix is multifold:
- Make the public `join` method only be used for comma-separation joining. Rename it to `joinWithCommaSeparator`.
- Wrap each group in `or` and `and` in parentheses to ensure correct precedence.
    ```
    > SQLFragment.or(sql`account_id = ${accountId}`, sql`app_id = ${appId}`
    sql`(account_id = ${accountId}) OR (app_id = ${appId})`
    ```

# Test Plan

Run existing tests to ensure nothing breaks, update unit tests and inspect output SQL statements to ensure they are correct parentheses. Especially the `builds complex queries with multiple helpers` test case.

PR #475, opened 2026-02-26
feat: add method to get pagination cursor for single entity

wschurman:
# Why

In a decent number of tests in the Expo server application, it's useful to test pagination from a specific point. It may also be useful for production code in an thus far unbeknownst way.

To do this, one must be able to generate the cursor given an entity. This PR adds such a method.

# How

Add the method that creates a cursor.

# Test Plan

Run the new test.

PR #474, opened 2026-02-26
fix: correct discriminated union type for forward/backward pagination

wschurman:
# Why

When testing this downstream, I noticed that it was allowed in types to specify first & before / last & after which aren't supported in the pagination.

# How

Fix the types by being explicit about the other fields.

# Test Plan

`yarn tsc`

PR #472, opened 2026-02-26
fix: Apply same viewer context subclass type mechanism to knexLoader and related methods

wschurman:
# Why

When these were installed extension functions, we could get away without the `TMViewerContext2` technique. But now that these are loader methods on the `PostgresEntity` subclass, we need this again (it's the same as Entity and ReadonlyEntity).

This type allows the typechecker to behave correctly when a subclass of ViewerContext is used.

# How

Add generic the same as ReadonlyEntity/Entity.

# Test Plan

`yarn tsc` in application that uses ViewerContext subclass and knexLoader.

PR #477, opened 2026-02-26
fix: correct pagination behavior for standard pagination with descending ordering clauses

wschurman:
# Why

This bug was subtle but important, and can be demonstrated best by looking at the new test case added `performs forward pagination with descending order`.

The bug was that cursor comparison must be derived from both direction but also ASC/DESC of the order bys.

Example:
```
obj1(t=1)
obj2(t=2)
obj3(t=3)
obj4(t=4)

> loadPage(first: 2, orderBy: t DESC)
obj4
obj3

> loadPage(first: 2, orderBy: t DESC, after: obj3)
```

the issue is that the cursor comparison for this second load would be:
```
(t) > (obj3.t)
```
which yields only obj4, which is incorrect.

# How

The fix is to determine cursor comparator by both the order by direction and the pagination direction, thus making the cursor from the example above:
```
(t) < (obj3.t)
```

The second part of this is to require all orderBys to have the same ordering for pagination, otherwise this new assumption breaks since cursor comparison then would be unable to be done with a single comparator. While there are ways we can handle different order by directions, we'd have to not use a cursor which adds a massive amount of complexity.

A small other fix was to set the `id` tiebreaker column order to the same as the other orders.

# Test Plan

Add new test case.

### DIFF
--- a/packages/entity-database-adapter-knex/src/__integration-tests__/PostgresEntityIntegration-test.ts
+++ b/packages/entity-database-adapter-knex/src/__integration-tests__/PostgresEntityIntegration-test.ts
@@ -1641,22 +1641,22 @@ describe('postgres entity integration', () => {
               strategy: PaginationStrategy.STANDARD,
               orderBy: [
                 { fieldName: 'hasACat', order: OrderByOrdering.DESCENDING }, // true comes before false
-                { fieldName: 'name', order: OrderByOrdering.ASCENDING },
-                { fieldName: 'id', order: OrderByOrdering.ASCENDING },
+                { fieldName: 'name', order: OrderByOrdering.DESCENDING },
+                { fieldName: 'id', order: OrderByOrdering.DESCENDING },
               ],
             },
           });
 
-          // Entities with cats (true) come first, then sorted by name
+          // Entities with cats (true) come first, then sorted by name descending
           expect(page.edges).toHaveLength(4);
           expect(page.edges[0]?.node.getField('hasACat')).toBe(true);
-          expect(page.edges[0]?.node.getField('name')).toBe('Alice');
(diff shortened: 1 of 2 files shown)


wschurman:
* **#477** <a href="https://app.graphite.com/github/pr/expo/entity/477?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/entity/477?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#476** <a href="https://app.graphite.com/github/pr/expo/entity/476?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#475** <a href="https://app.graphite.com/github/pr/expo/entity/475?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#474** <a href="https://app.graphite.com/github/pr/expo/entity/474?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#470** <a href="https://app.graphite.com/github/pr/expo/entity/470?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#472** <a href="https://app.graphite.com/github/pr/expo/entity/472?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

wschurman:
### Merge activity

* **Feb 27, 2:15 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/expo/entity/477).
* **Feb 27, 2:32 AM UTC**: [Graphite](https://app.graphite.com/github/pr/expo/entity/477) rebased this pull request as part of a merge.
* **Feb 27, 2:35 AM UTC**: @wschurman merged this pull request with [Graphite](https://app.graphite.com/github/pr/expo/entity/477).

PR #506, opened 2026-03-05
feat: Add SQLExpression class that supports fluent comparison method API

wschurman:
# Why

When converting the Expo server's `loadManyByRawWhereClause` to `loadManyBySQL` I found myself wanting the ability to do something like `jsonPath('fieldName', 'wat').isNull()` rather than having to use the template tag.

# How

Add SQLExpression with claude to support a better API for expressions, and add a few more common methods as well.

# Test Plan

Run all tests.

PR #516, opened 2026-03-09
feat: update the SQLExpression and SQLFragmentHelpers APIs

wschurman:
# Why

The `expression` API added in #506 was somewhat poisonous. When doing something like `jsonPath('fieldName', 'wat').eq(10)`, the type information was lost due to typescript not carrying it through chains. So I could do something like:
```
BlahEntity.knexLoader()..loadManyBySQL(sql`${entityField('hadsACat')} = ${true}`)
```
and it would correctly tell me that the field name is invalid (type info carries into single arg).

But the following wouldn't:
```
BlahEntity.knexLoader().loadBySQL(expression('hadsACat').eq(true))
```

# How

It's better to just get rid of `expression` though keep the fluent API on SQLFragmentHelper methods that can carry the type information through the chain.

This PR does a few things:
- adds the required code to carry that type information through
- updates SQLFragmentHelper methods to take either an expression/fragment or entity field name for better usability.
- renames SQLExpression to SQLChainableFragment to better describe what it does
- renames SQLFragmentHelper to SQLExpression to better describe the class of things it contains.

# Test Plan

Run new tests.

### DIFF
--- a/packages/entity-database-adapter-knex/src/EnforcingKnexEntityLoader.ts
+++ b/packages/entity-database-adapter-knex/src/EnforcingKnexEntityLoader.ts
@@ -159,7 +159,7 @@ export class EnforcingKnexEntityLoader<
    *   .limit(10)
    *   .executeAsync();
    *
-   * const { between, inArray } = SQLFragmentHelpers;
+   * const { between, inArray } = SQLExpression;
    * const filtered = await ExampleEntity.loader(vc)
    *   .loadManyBySQL(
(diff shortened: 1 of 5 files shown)


wschurman:
* **#516** <a href="https://app.graphite.com/github/pr/expo/entity/516?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/entity/516?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #540, opened 2026-03-20
feat: add `updateWithoutReloadingAsync` to skip `RETURNING *` and post-update reload

ide:
## Summary

`updateAsync` always generates `UPDATE ... RETURNING *`, then issues a second `SELECT` to reload the entity for after-triggers and to return to the caller. For fire-and-forget writes where the caller discards the result, both the `RETURNING *` clause and the reload SELECT are pure waste. This is one fewer database round trip per call.

The database adapter layer gains `updateWithoutReturningAsync`, which omits `RETURNING *` and returns the affected row count instead of the full row. Row-existence validation uses the count (0 means the row was deleted concurrently, same error as before).

The mutator layer gains `updateWithoutReloadingAsync` on both `AuthorizationResultBasedUpdateMutator` (returns `Result<void>`) and `EnforcingEntityUpdater` (returns `Promise<void>`). The naming reflects each layer's abstraction: the adapter speaks in terms of SQL's `RETURNING` clause, while the mutator speaks in terms of the entity reload.

Authorization, validators, before-triggers, and cache invalidation still run. The method throws immediately — before entering a transaction — if the entity has `afterUpdate`, `afterAll`, or `afterCommit` triggers, since silently skipping them would be dangerous.

## Test plan

- `yarn tsc && yarn lint && yarn test`

### DIFF
--- a/packages/entity-database-adapter-knex-testing-utils/src/StubPostgresDatabaseAdapter.ts
+++ b/packages/entity-database-adapter-knex-testing-utils/src/StubPostgresDatabaseAdapter.ts
@@ -278,6 +278,34 @@ export class StubPostgresDatabaseAdapter<
     return [objectCollection[objectIndex]];
   }
 
+  protected async updateWithoutReturningInternalAsync(
+    _queryInterface: any,
+    tableName: string,
+    tableIdField: string,
+    id: any,
+    object: object,
+  ): Promise<number> {
+    if (Object.keys(object).length === 0) {
+      throw new Error(`Empty update (${tableIdField} = ${id})`);
+    }
+
+    const objectCollection = this.getObjectCollectionForTable(tableName);
+
+    const objectIndex = objectCollection.findIndex((obj) => {
+      return obj[tableIdField] === id;
+    });
+
+    if (objectIndex < 0) {
+      return 0;
+    }
+
+    objectCollection[objectIndex] = {
+      ...objectCollection[objectIndex],
+      ...object,
+    };
+    return 1;
+  }
+
   protected async deleteInternalAsync(
     _queryInterface: any,
     tableName: string,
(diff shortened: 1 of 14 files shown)


ide:
I think profiling the change is the one of the most useful things we can do. The performance stats (Postgres, JS, Redis) would make the decision clearer.

The use case that motivated this PR was to optimize updating fields like `apps.latest_activity_time`, specifically the Postgres load. Removing `RETURNING *` would speed up the query due to less I/O and the row-level lock would be released a tad sooner. I am really interested to see how removing `RETURNING *` affects actual performance.

My concern about reloading the entity is I suspect we perform many writes to `apps.latest_activity_time` at the same time, triggering a lot of reloads. We have some logic to throttle these writes but there is still room for optimization, like coalescing the `latest_activity_time` triggers with a Batcher, for instance.

Another area of optimization at Postgres's level would be to support custom SQL expressions in entity updates, specifically the `WHERE` predicates. We could do something like:
```sql
WHERE id = :id AND
  -- Custom WHERE predicate
  (latest_activity_time IS NULL OR latest_activity_time < :time - interval '60 seconds')
```
for instance.

I propose:
1. Deploy your `RETURNING *` removal.
2. Explore `WHERE` predicates, along with an optimization to skip reloading an entity if the database skipped the update (updated row count is zero).
3. Explore the implications of batching in triggers (not generalized, just for one or two triggers).

If we have all three of these implemented, this PR is entirely obviated I think.

wschurman:
> Explore WHERE predicates, along with an optimization to skip reloading an entity if the database skipped the update (updated row count is zero).

Agreed 100%. Now that knexLoader is split out, I think it also makes sense to have `knexCreator`/`knexUpdater`/`knexDeleter` that offer APIs specific to knex/postgres. I'd previously begun to explore batch mutation in #446 and batch deletion in #554, both of which I was first trying to make general enough to live outside of knex mutators, but thinking more it probably makes sense for these to be in knex mutators since they'll likely require postgres-specific sql to be useful.

Let's explore creating a knex mutator with an additional where predicate just to see how it looks, following the same knex-specific extension pattern to add the new mutator class.

wschurman:
Exploring with claude in #558.

ide:
Closing for now; we can always re-open if wanted.

PR #555, opened 2026-04-01
fix: remove unnecessary RETURNING * from database adapter update

wschurman:
# Why

When reviewing #540, I noticed that the `RETURNING *` isn't necessary, even in the normal case (since #233).

# How

Remove the `RETURNING *`, return a simple updatedRowCount instead for error behavior.

# Test Plan

Run tests.


wschurman:
* **#555** <a href="https://app.graphite.com/github/pr/expo/entity/555?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/entity/555?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #565, opened 2026-04-06
feat: add count equivalent for knex loader methods

wschurman:
# Why

Looking at more ways the Expo server application needs escape hatches from entity, particularly in postgres/knex, it's:
- count
- upsert
- batch create/update/delete (and then batch count/upsert/etc)

In other PRs I'm looking into batch stuff, but it's quite complex. In the meantime, some easy wins are count and upsert.

# How

Adds count equivalents for knex loads. Note that these purposefully don't do authorization since the point of these is to not need to load the full row set, and authorization is row-based.

# Test Plan

Run tests.

### DIFF
--- a/packages/entity-database-adapter-knex/src/PostgresEntityDatabaseAdapter.ts
+++ b/packages/entity-database-adapter-knex/src/PostgresEntityDatabaseAdapter.ts
@@ -1,5 +1,5 @@
 import type { EntityConfiguration, FieldTransformer, FieldTransformerMap } from '@expo/entity';
-import { getDatabaseFieldForEntityField } from '@expo/entity';
+import { getDatabaseFieldForEntityField, RESERVED_ENTITY_COUNT_QUERY_ALIAS } from '@expo/entity';
 import type { Knex } from 'knex';
 
 import type {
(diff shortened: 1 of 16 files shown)


wschurman:
* **#566** <a href="https://app.graphite.com/github/pr/expo/entity/566?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#565** <a href="https://app.graphite.com/github/pr/expo/entity/565?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/entity/565?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ide:
For upsert, there is now MERGE in Postgres. Awhile ago Sonnet 4.6 claimed it was not as efficient as ON CONFLICT but I think it would be best to profile both in a production scenario and see if we can tell a difference in the Grafana query stats.

packages/entity-database-adapter-knex/src/PostgresEntityDatabaseAdapter.ts
```diff
@@ -206,29 +204,79 @@ export class PostgresEntityDatabaseAdapter<
...
+    queryInterface: Knex,
+    tableName: string,
+    tableFieldSingleValueEqualityOperands: TableFieldSingleValueEqualityCondition[],
+    tableFieldMultiValueEqualityOperands: TableFieldMultiValueEqualityCondition[],
+  ): Promise<number> {
+    const query = this.applyFieldEqualityConjunctionWhereClause(
+      queryInterface.count('* as count').from(tableName),
+      tableFieldSingleValueEqualityOperands,
+      tableFieldMultiValueEqualityOperands,
+    );
```
ide: This is subtle but worth adding a test for IMO: if we have a WHERE clause that tries to match on `count = ?`, this `count` refers to a column named `count` in the table - not the `count(*). (If no such column exists, PG fails)

ide (approved):

quinlanj (approved):

ide:
Another concern I have is performance, specifically big scans taking a long time. Docblocks that emphasize this seem important.

wschurman:
Added a docblock to emphasize that indexes are just as critical for these count methods as they are for their non-count counterparts.

packages/entity-database-adapter-knex/src/PostgresEntityDatabaseAdapter.ts
wschurman: Nice catch! Renamed the count to `__entity_count__` and marked it as reserved.

PR #582, opened 2026-04-13
fix: upgrade knex and remove undefined value bindings

wschurman:
# Why

Renovate upgraded knex, which exposed one constraint that wasn't previously expressed in the knex types: undefined, while it does technically work (I think? maybe not), is not supported as a value binding.

It was added to a lot of methods to simplify the expressions: https://github.com/expo/entity/pull/414#pullrequestreview-3749427513
I think it was originally added to SupportedSQLValue by claude.
It's not necessary since entity field values should never be null, so they shouldn't be compared to null, and undefined isn't a postgres value, so in general it shouldn't be in the entity sql dialect.

This was exposed by https://github.com/knex/knex/pull/5373.

# How

1. Upgrade knex, remove undefined value support from a number of places.
1. Simplify some logic in places that used to implicitly supply undefined values, mostly due to unchecked array accesses.

# Test Plan

Run all tests.

### DIFF
--- a/packages/entity-database-adapter-knex/package.json
+++ b/packages/entity-database-adapter-knex/package.json
@@ -27,7 +27,7 @@
   },
   "dependencies": {
     "@expo/entity": "workspace:^",
-    "knex": "^3.1.0"
+    "knex": "^3.2.9"
   },
   "devDependencies": {
     "@expo/entity-testing-utils": "workspace:^",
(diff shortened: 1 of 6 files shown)


wschurman:
* **#581** <a href="https://app.graphite.com/github/pr/expo/entity/581?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#582** <a href="https://app.graphite.com/github/pr/expo/entity/582?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/entity/582?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

quinlanj (approved):

PR #628, opened 2026-05-28
fix: require transactional query context for createOrGetExistingAsync utility method

wschurman:
# Why

One class of issue in entity is concurrent calls to load/create methods caching stale data in dataloaders/redis when done outside a transaction (otherwise dataloaders are scoped to transaction and cache adapter is skipped). This can happen when multiple calls to blocks of code like `createOrGetExistingAsync` happen in parallel without an explicit transactional query context, within which it would negatively cache in one call and then incorrectly retrieve that negative cache from another call depending on race condition ordering.

Concretely, one such issue with `createOrGetExistingAsync` is reproducible by calling it with the same set of args:
```typescript
const [a, b] = await Promise.all([
  createOrGetExistingAsync(...args),
  createOrGetExistingAsync(...args),
]);
```
This will throw due to the ordering: both calls will miss initially, both will create, one will hit unique constraint error (handled), that one will try to re-load, but will see the negatively-cached result in dataloader.

# How

While the class of bug can happen with really any non-transactional sequence of loads/mutations, entity as a library should do its best to prevent it from occurring in blessed utility methods. The only way to make this deterministic is via requiring a transactional query context be supplied to `createOrGetExistingAsync`.

Reasoning for why this is likely the right call path. Most creations in applications should already be within transactions, and those that aren't automatically start a transaction in the mutator to run the triggers and create within the same transaction. This change makes the semantics of this method more explicit without much loss of usability and protects against the race that is likely.

# Test Plan

ToDo

### DIFF
--- a/packages/entity-database-adapter-knex/src/__integration-tests__/EntityCreationUtils-test.ts
+++ b/packages/entity-database-adapter-knex/src/__integration-tests__/EntityCreationUtils-test.ts
@@ -33,57 +33,6 @@ describe(createWithUniqueConstraintRecoveryAsync, () => {
   });
 
   describe.each([true, false])('is parallel creations %p', (parallel) => {
-    it('recovers when the same entity is created twice outside of transaction', async () => {
-      const vc1 = new ViewerContext(createKnexIntegrationTestEntityCompanionProvider(knexInstance));
-
-      const args = {
-        name: 'unique',
-      };
-
-      let createdEntities: [PostgresUniqueTestEntity, PostgresUniqueTestEntity];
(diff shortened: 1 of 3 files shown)


wschurman:
* **#628** <a href="https://app.graphite.com/github/pr/expo/entity/628?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/entity/628?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #656, opened 2026-07-14
Add resolveGrantDecision permission-resolution helper

tomek-agent:
# Why

A small reusable helper for a common authorization shape: deciding whether a subject holds a permission on a resource that lives in a containment hierarchy (resource → parent → root), with an optional per-resource opt-out from inheritance.

# How

Adds `resolveGrantDecision`, a pure, dependency-free function. The caller gathers the data (the resource's ancestry, the permissions held keyed by resource, and whether the target opts out of inheritance for the required permission) and this function applies the resolution rule. Keeping the rule separate from data gathering keeps it easy to unit-test and to reuse across different storage and permission models.

# Test Plan

Unit tests cover direct grants, inherited grants, a non-matching permission, the inheritance opt-out (both directions), and the empty-ancestry case.

### DIFF
--- a/packages/entity/src/GrantResolution.ts
+++ b/packages/entity/src/GrantResolution.ts
@@ -1,0 +1,66 @@
+/**
+ * A small, pure helper for resolving whether a subject holds a required permission over a
+ * resource that lives in a containment hierarchy (e.g. resource -> parent -> root).
+ *
+ * It is intentionally dependency-free and total: the caller supplies the already-gathered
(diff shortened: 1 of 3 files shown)


tomek-agent:
Closing this. On reflection the helper is decoupled from the framework's abstractions and doesn't earn a public API here; folding it into the consumer for now, and will upstream a genuinely framework-shaped primitive later if one emerges.

PR #662, opened 2026-07-22
feat!: emit a dedicated metric event for count queries

gwdp:
# Why

Count queries reported their result through the load event, so a `COUNT(*)` returning 8k looked like we loaded 8k entities.

# How

Give counts their own event (`logDataManagerCountEvent`) so the count number stops being read as entities loaded. Breaking: `IEntityMetricsAdapter` implementers need to add the method; I'm happy to go for another path, non-breaking change if you prefer, but this felt more fundamentally correct 

giphy

# Test Plan

CI. Updated the metrics unit test to check the counts route to the new event.


wschurman (approved):
[awesomesauce](https://app.graphite.com/settings/meme-library?org=expo)

PR #672, opened 2026-08-03
chore: treat typedoc warnings as errors to keep evergreen

wschurman:
# Why

#671 fixed some. This changes the CI so that we don't have to periodically fix these. Instead, CI will fail on warnings.

# How

Add setting.

# Test Plan

CI


gwdp (approved):
🔥

PR #699, opened 2026-09-17
feat!: add user-facing denial reason to privacy policies

wschurman:
# Why

One product request at Expo is the ability to use authorization as a product feature. 

For example, let's say an account has a flag, `require_higher_permission`, that when enabled, requires a higher user permission to view some entity owned by that account. The way this should be implemented is in privacy policies: in a rule, check if the account has the flag, and if so, require a higher permission in the read policy for the entity in question, only `RuleEvaluationResult.ALLOW`ing when the user has been granted the higher permission.

The issue is that if all rules skip, a denial will be issued but without sufficient information about why. So the product won't be able to tell the end user that they need a higher permission since the boolean is enabled.

# How

This PR fixes this by adding an optional "reason" system on top of privacy policies. It works as follows:
1. Each time a rule skips or denies, it can provide a reason code. These are accumulated during the policy execution into a skip reason code list.
1. At the end of privacy policy evaluation, when the `EntityNotAuthorizedError` is constructed and thrown, a hook is provided (`getUserFacingDenialReason`) on the policy class for the developer to optionally translate the list of skip reason codes into a better human-readable error message, which is attached to the `EntityNotAuthorizedError`.
1. In a consumer application, in the application error formatter, it can check if the error is a `EntityNotAuthorizedError`, and if so, it can choose how to use the human-readable error message.

This also updates the privacy policy testing helpers to support code assertion.

# Test Plan

Run updated tests.


wschurman:
* **#699** <a href="https://app.graphite.com/github/pr/expo/entity/699?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/entity/699?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/entity/src/errors/EntityNotAuthorizedError.ts
```diff
@@ -3,6 +3,70 @@ import type { ReadonlyEntity } from '../ReadonlyEntity.ts';
...
+ */
+export type EntityNotAuthorizedDenialReason =
+  | {
+      type: EntityNotAuthorizedDenialType.RULE_DENIED;
+      rule: EntityNotAuthorizedRuleInfo;
+    }
+  | {
+      type: EntityNotAuthorizedDenialType.ALL_RULES_SKIPPED;
+      skippedRules: readonly EntityNotAuthorizedRuleInfo[];
+    };
```
byronkarlen, line 59: Probably a niche scenario, but in the case that a rule gets denied, don't we also want to show `EntityNotAuthorizedRuleInfo` for any skipped rules that happened before it?

packages/entity/src/errors/EntityNotAuthorizedError.ts
```diff
@@ -26,16 +90,43 @@ export class EntityNotAuthorizedError<
...
+   * and facts about the entity and viewer; it is intentionally excluded from `message` and must
+   * not be surfaced directly to end users.
+   */
+  public readonly denialReason: EntityNotAuthorizedDenialReason;
+
+  /**
+   * End-user-safe explanation supplied by the privacy policy, if any.
+   * See EntityPrivacyPolicy.getUserFacingDenialReason.
+   */
+  public readonly userFacingReason: EntityNotAuthorizedUserFacingReason | null;
```
byronkarlen: nit: thoughts on calling it `userFacingDenialReason` instead? Also stays consistent with `getuserFacingDenialReason()`

packages/entity/src/EntityPrivacyPolicy.ts
```diff
@@ -501,11 +539,25 @@ export abstract class EntityPrivacyPolicy<
...
+    });
+  }
+
+  private createNotAuthorizedError(
+    viewerContext: TViewerContext,
+    entity: TEntity,
+    action: EntityAuthorizationAction,
+    denialReason: EntityNotAuthorizedDenialReason,
+  ): EntityNotAuthorizedError<TFields, TIDField, TViewerContext, TEntity, TSelectedFields> {
+    const userFacingReason = this.getUserFacingDenialReason(viewerContext, action, denialReason);
```
byronkarlen, line 555: what if this throws? Should we be handling this?

byronkarlen:
Also need to document that this is a breaking change due to `EntityNotAuthorizedError` api(?)

packages/entity/src/EntityPrivacyPolicy.ts
wschurman: Generally the entity library doesn't try to protect against consumer errors within the integration points, and also doesn't gracefully handle them. For example, if an entity throws in `defineCompanionDefinition`, the error is just propagated through the call stack as normal.

The one exception to this is in [Entity construction](https://github.com/expo/entity/blob/main/packages/entity/src/EntityConstructionUtils.ts#L162) where we explicitly wrap it in a result so that consumers can override the constructor to do last-minute data validation and still use the `withAuthorizationResults` API format. Note that the `enforcing` API format would still throw though.

packages/entity/src/errors/EntityNotAuthorizedError.ts
wschurman: Good point! Added

packages/entity/src/EntityPrivacyPolicy.ts
byronkarlen: Gotcha good to know

wschurman:
Re: breaking change - because the API changes to EntityNotAuthorizedError are only additive, I think it's backwards-compatible. Main error message also should remain unchanged.

byronkarlen:
> because the API changes to EntityNotAuthorizedError are only additive

We got rid of the `ruleIndex` param though?

wschurman:
Ah, I see what you mean. Yep, technically breaking. Will update!

wschurman:
(making it breaking meant just adding a `!` to the commit/PR lol)

byronkarlen:
ah nice

byronkarlen (approved):

